PR #15, opened 2025-01-30
Hopper Team Agreement and Write-Up 

garciaow:


EvilJuiceBox (changes requested):
Please remove all the other cpp files before you stage the commit. The first commit should only contain the writeup and agreements

PR #16, opened 2025-01-30
[WIP] [Group 4] BitVector class

CircleShift:
Pull request in preparation for when we finish the BitVector class


CircleShift:
The BitVector class now has an implementation of everything I would expect from a basic version of this class:
- Indexing
- Bitwise AND, OR, XOR, and NOT
- Bit shift operations
- Equality
- Printing to stream
- Resize operation, with multiple fill options
- Append operation
- Prepend operation
Please note that while initial implementations of the above are provided, they may not be correct as testing still needs to be worked on.

CircleShift:
Please use this commit for testing b580226f847c04d38a5633c10a76bba8d6c02d11

CircleShift:
See #137

PR #18, opened 2025-01-31
Hamilton - Datum Class 

MaxKrawec:
Created initial Datum Class. Methods implemented:
- Constructors
- AsString/Double
- IsString/Double
- SetStringValue
- SetDoubleValue
- GetVariant
- GetString/Double

Also implemented:
- If an invalid string tries to convert into a double, returns NaN
- If NaN (double) tries to convert into a string, returns an empty string

TODO:
- Implement  operations (+,-,*,/,>,<,=, etc)
- Better testing/test framework

Notes:
1) Since we don't have a testing framework yet, I added some very basic tests to main (Felt weird to push code without testing). Will need to move and improve the tests once we pick a testing framework. Also, probably remove main at some point (I think).

2) I'm not sure how we want to set up/organize the project. I just added a few folders to help organize it a bit. Can change later

3) I think adding the CMakeLists.txt is okay. I'm trying to use CLion for the project, and I think it needs it (could be wrong). If that causes any issues let me know.

### DIFF
--- a/Group-05/CMakeLists.txt
+++ b/Group-05/CMakeLists.txt
@@ -1,0 +1,6 @@
+cmake_minimum_required(VERSION 3.30)
+project(hamilton)
+
+set(CMAKE_CXX_STANDARD 20)
+
+add_executable(hamilton main.cpp src/Datum.h src/Datum.cpp)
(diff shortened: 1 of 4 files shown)


MaxKrawec:
If the professor/TA sees this, this is a PR to pull this branch into our dev branch. You don't have to review.

PR #23, opened 2025-02-03
Implemented DataGrid class, member functions, and simple tests

ShahaabAli:


Group-05/src/DataGrid.cpp
MaxKrawec, line 1: Mixed between snake_case and camelCase for common variables. Google format wants snake_case. Examples:
rowIndex, columnIndex, etc.

Group-05/src/DataGrid.h
MaxKrawec, line 1: Same comment about snake_case: Mixed between snake_case and camelCase for common variables. Google format wants snake_case. Examples:
rowIndex, columnIndex, etc.

Group-05/src/DataGrid.h
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Index a row of the DataGrid
+   * @param rowIndex
+   * @return indexed vector row from DataGrid
+   */
+  std::vector<Datum> &GetRow(const std::size_t rowIndex) {
+    return mGrid.at(rowIndex);
+  }
+
+  std::vector<Datum> GetColumn(std::size_t columnIndex) const;
```
MaxKrawec, line 57: Should this return a reference like GetRow does? AKA: std::vector< Datum > &GetColumn...

Group-05/src/DataGrid.h
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Get Datum value from DataGrid
+   * @param rowIndex
+   * @param columnIndex
+   * @return the Datum at the given index
+   */
+  Datum &GetValue(const std::size_t rowIndex, const std::size_t columnIndex) {
+    return mGrid.at(rowIndex).at(columnIndex);
+  }
+
```
MaxKrawec, line 68: Is the GetValue method suppose to be here twice?

Group-05/src/DataGrid.cpp
```diff
@@ -0,0 +1,133 @@
...
+namespace cse {
+
+/**
+ * Returns the shape of the DataGrid
+ * @return tuple of (# of rows, # of columns)
+ */
+std::tuple<const std::size_t, const std::size_t> DataGrid::Shape() const {
+
+  const std::size_t num_of_rows = mGrid.size();
+
```
MaxKrawec, line 22: Note: I don't understand the code below determining the number of columns. But if you do, and tested it, we should be fine.

MaxKrawec (commented):
Overall it looks good. A few things:
1) Consistent use of snake_case for common variables
2) Should it be a reference return type for GetColumn?
3) It looks like GetValue is done twice, I don't fully understand that
4) I don't understand the number of columns code. But if you do, and tested it, we should be fine.

Group-05/src/DataGrid.h
ShahaabAli: There is a bug when returning a reference. Should be fixed when we use a ReferenceVector instead.

Group-05/src/DataGrid.h
ShahaabAli: Used for printing method using streams

Group-05/src/DataGrid.cpp
ShahaabAli: Utilizing an algorithm to find the maximum number of values in a row

Group-05/src/DataGrid.cpp
ShahaabAli: Will update soon

Group-05/src/DataGrid.h
ShahaabAli: Will update soon

PR #24, opened 2025-02-03
Added Event, EventQueue classes. Implemented initial logic

owenhaiar:


graceKC04 (approved):
I think the EventCompare sign may just need to be switched but otherwise it looks good. It also may be useful to considering switching to .hpp instead to indicate it being a C++ file.

allienello (approved):
Looks good!

graceKC04 (approved):

PR #26, opened 2025-02-04
Hamilton parser

PMitkiewicz:


MaxKrawec (changes requested):
You need to fix your branch. In your branch, you have everyone's code and files. It should just be your own code

asifmasood7:
@PMitkiewicz Hi Pedro, I had that same issue where everyone's commits were showing into my branch as well and I was doing it correctly in Github, but was making a mistake in VS code when pushing my changes. (Kira and professor said that's normal but I wanted to take another close look into it this evening). Make sure to be in hamilton branch in vs code (not main or any other branch) and then create your own parser branch, and after that, create your own PR. That wored for me so wanted to update you on that since we both had same issue.

PMitkiewicz:
Got it, thanks!

PR #31, opened 2025-02-05
Hamilton - Datum Features 

MaxKrawec:
Completed:
1) Added Datum overloaded functions for +, -, *, /, ==. 
2) Fixed trailing 0s for double to string conversion.

Notes:
1) For +, -, *, /, it calculates the value for two Datum doubles. Otherwise, returns NaN (Two Strings or one double and one string) 2) For ==, it works for two double Datums or two string Datums. Also, due to double imprecision, needed an epsilon. I picked a random, small number. May want to choose a different one. 3) Didn't fully test all the functions in main.cpp yet. Going to wait for our testing framework.

TODO:
1) Need to discuss how we want to implement >, >=, <, <=. Do we mix and match strings and doubles? Do we separate them? I don't know what's best. 2) Add tests when we pick a testing framework

### DIFF
--- a/Group-05/main.cpp
+++ b/Group-05/main.cpp
@@ -71,6 +71,30 @@ int main() {
   std::cout << "NaN to String: Should have an empty string: "
             << d6.GetString().value() << std::endl;
 
+
+  cse::Datum d_add(28.421);
+  cse::Datum d_add1(38.32);
+  cse::Datum d_add2("test");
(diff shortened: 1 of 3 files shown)


MaxKrawec:
Note: Probably want to wait for this PR. I didn't include that many tests in main.cpp because I want to wait for our testing framework

PR #36, opened 2025-02-05
feat: reference vector class implementation and testing

greencalen17:


Group-05/src/ReferenceVector.h
```diff
@@ -0,0 +1,115 @@
...
+   private:
+    /// Stores a pointer to a Datum.
+    Datum* mDatum;
+
+   public:
+    /**
+    * Constructs a Proxy object for a given Datum pointer.
+    *
+    * @param datum A pointer to the Datum instance.
+    */
```
MaxKrawec, line 37: In the professor comments in the discord, he says " I think you are correct that you don't need to worry about the proxy objects at this time; "

To be honest, I don't fully understand the proxy class code. Personally, I would rewrite it without the proxy, but if you understand it, maybe it's fine?

Group-05/src/ReferenceVector.h
MaxKrawec, line 1: I would implement the `void SetReference(size_t index, Datum & var)` the professor recommended. 

Also, I wrote this in the Discord about ReferenceVector, just rewriting it here so you can see it. Let me know if you have any questions about it:

In addition to SetReference, I think we should also have PushBack(Datum & var) to add values. If I understand SetReference correctly, it's basically just sets values in the vector member variable. But if the ReferenceVector starts off empty, we need a way to add values (With an empty vector SetReference(0, datum1) should cause undefiend behavior, right?). Instead, we should call the new PushBack method first which calls the vector push_back method on the member variable. Now, you can run SetReference(0, datum1) and it shouldn't cause undefined behavior. 

Then we could add all the useful vector methods. I know we thought we could potentially do this with a proxy class, but I'm not if we can. Also, I'm pretty sure Ofria said it isn't needed. So we can just make the similar vector functions ourselves. Let me know if any of this didn't make sense (we can also talk in class)

MaxKrawec (commented):
Overall, looks good. Few comments:
1) I didn't fully understand the Proxy class. Since the professor said "I think you are correct that you don't need to worry about the proxy objects at this time, they would only be needed for more esoteric functionality." I think I would re-write the class without it
2) I would add the `void SetReference(size_t index, Datum & var)` when you can. I left detailed notes in one of the comments about this.  
3) Eventually, (Probably don't need it for friday) need to make the member variable be a template type. 
4) Also something we don't need by Friday, but maybe add more type functions.

PR #38, opened 2025-02-05
Completed initial  working classes for specs

gr005100:
Initial working C++ classes for class random

### DIFF
--- a/Group-01/Random/Random.hpp
+++ b/Group-01/Random/Random.hpp
@@ -1,0 +1,47 @@
+#ifndef RANDOM_HPP
+#define RANDOM_HPP
+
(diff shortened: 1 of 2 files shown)


KollinBartz (approved):
Looks good

Group-01/Random/Random.hpp
```diff
@@ -0,0 +1,47 @@
+#ifndef RANDOM_HPP
```
LeoSSpecht, line 1: #pragma once

LeoSSpecht (approved):

LeoSSpecht (changes requested):
#pragma once

PR #53, opened 2025-02-07
added TextBox and FormattedText classes and test files

allienello:
Added TextBox and FormattedText classes, and implemented unit tests for both classes.


merryymary (approved):
All the tests passed when I used it on my end, so looks all good!

PR #63, opened 2025-02-07
[Lamport]-[Group-6] Final versions of working classes with Advanced specs and tests

matrix373:
We have implemented initial working versions of each class and created a Doxygen configuration with Makefile for our project.


matrix373:
@EvilJuiceBox  @mercere99

IvanBega:
Merged tests and asserts into Lamport

matrix373:
Final Versions of Initial classes for Lamport

nstanis21 (commented):
Review for StringSet:
You have quite a few unnecessary comments that are just polluting the codespace since they are describing code which is on its own already very readable, some examples are lines 15, 19, 24, 37, 40, 42-44, 59, 64, 80, 87, 100, and 107 in StringSet.cpp. I feel that the code would be more readable without these comments.

The Filter and FilterOut functions have so much overlap, could they be combined into a single Filter function that takes a filterin/filterout parameter to determine if items are being filtered in or out (or just keep only the Filter function in its current form since Filter could also function as FilterOut with an inverted boolean function result)?

You sometimes had unnecessary/redundant tests in the testing space, it would make your tests much more readable if you got rid of them, some examples are lines 34, 65, and any instances of REQUIRE(set.size()==0) followed by REQUIRE(set.empty()) (such as in lines 150-155). Since you already have done the empty tests you shouldn't need to double up on these tests, and I feel it would make your tests much more readable.

Group-06/StaticString/StaticString.hpp
```diff
@@ -0,0 +1,588 @@
...
+  * @brief Index operator to get value in the specific index of string.
+  *
+  * Index operator only used to get a copy of character on the given index.
+  *
+  * @see set() to change a character on a given index.
+  *
+  * @param index size_t The index of character.
+  * @return Indexed char value.
+  */
+  char operator[](std::size_t index) {
```
sidb70, line 187: i would suggest adding a const version of this as well

sidb70 (commented):
Design

Informative comments. Comments are well placed and informative. 

Implementation
consider using std::array instead of a raw char array to store the characters. this will enable you to use the built-in array methods. 

instead of overloading append for 3 different types, consider making a single function which takes a begin and end iterator as parameters. 


Testing: current tests are comprehensive and cover all member functions. 
Edge cases/Error conditions: there is currently no test to check what happens when an out of bounds index is accessed. 

There is also currently no test for what happens when a user tries to use the substr() function with the end index being lower than the start index.

you may also consider adding tests that verify the const correctness of your code. for example, testing the cost version of get_str

rodriguezjoanna (commented):
**DESIGN**
_Informative Identifiers_
      •	Strength: Class name ArgManager and methods (Has, GetOption, GetOptions) are descriptive.
      •	Improvement: Consider more explicit names (e.g., HasArg, GetSingleOption).
_Avoids “Magic Numbers”_
      •	Strength: Code uses mArgc and loops accordingly. No hard-coded constants.
      •	Improvement: Not necessary here, no apparent arbitrary “magic” integers.
_Intuitive & Functional Interface_
      •	Strength: Straightforward constructor (argc, argv), easy method calls.
      •	Improvement: Could return std::optional for error cases instead of printing error messages.
_Informative Comments_
      •	Strength: Method descriptions are clear and succinct.
      •	Improvement: Consider focusing comments on why rather than what (e.g., error conditions, usage patterns).
_Consistent Style_
      •	Strength: mArgv and mArgc consistently named; braces and indentation look uniform.
      •	Improvement: If you adopt a project-wide style guide, ensure it’s documented (e.g., whether to use snake_case or camelCase for functions).
_Visual Appearance_
      •	Strength: Readable line lengths, good spacing.
      •	Improvement: None major; code is visually clean.

**IMPLEMENTATION**
_No Warnings During Compilation_
      •	Strength: Code compiles without warnings under -Wall -Wextra.
      •	Improvement: None.
_Proper Use of Namespaces_
      •	Strength: Everything is under namespace cse, no using namespace std; in headers.
      •	Improvement: None needed.
_Adheres to C++ Standard_
      •	Strength: Standard headers used, no platform-specific calls.
      •	Improvement: Could adopt modern return types (std::optional) for error states.
_Appropriate Modularity_
      •	Strength: Header/source separation is good; the class is focused on argument parsing.
      •	Improvement: Minor possibility: remove duplicate logic with a helper function to find argument indices.
_Uses Modern C++ Constructs_
      •	Strength: Use of std::vector<std::string>.
      •	Improvement: Employ range-based for loops or std::find instead of manual indexing.
_Uses Standard Library Algorithms_
      •	Strength: Basic iteration is fine.
      •	Improvement: Can use std::find to check for existence.
_Pointer Handling_
      •	Strength: Minimal pointer usage; copies argv[] into a safe std::vector.
      •	Improvement: None needed for dynamic allocation management.
_Const Correctness_
      •	Strength: Methods mostly read data.
      •	Improvement: Mark Has, GetOption, GetOptions as const; pass strings by const std::string&.

**TESTING AND ERROR CHECKING**
_Unit Testing_
Strengths
      •	Each public member function (Has, GetOption, GetOptions) is tested with at least one scenario.
      •	Tests are split into logical SECTIONs, making them readable and well organized.
      •	The constructor is implicitly tested each time ArgManager is created.
Improvement
      •	Zero arguments: You could add a test with argc = 0 to ensure the class handles empty command lines gracefully.
      •	Non-integer parsing: GetOptions internally calls std::stoi, but there’s no test to ensure that invalid numeric input (ex, a string like "foo") produces the expected behavior.
      •	Constructor edge cases: Confirm behavior if argc is negative (rare but could be tested just to confirm error handling).

_Edge Cases_
Strengths
You test:
      •	Missing arguments (DoesNotExist).
      •	Not enough arguments in GetOptions (tries to parse 7 but only 5 are provided).
      •	Single-argument vs. multiple-arguments for GetOption.
Improvement
      •	Zero or no arguments: Currently not tested.
      •	Non-integer for GetOptions: The function will catch an exception and print an error message, but your tests don’t confirm that behavior.
_Assertions_
Strengths
      •	The tests use REQUIRE from Catch2
      •	There are no explicit assert statements inside ArgManager or the tests themselves. Typically fine in high-level code, but using assert inside ArgManager can document assumptions (ex, “index must never exceed vector size”).
Improvement
      •	If there are invariants that must always hold (like a required minimum number of arguments), you can use assert within the class to catch programmer errors early.
      •	Since the code recovers from errors by returning "" or printing messages, asserts may not be strictly necessary but they are an option for internal consistency checks.
_Error Handling_
Strengths
      •	Tests confirm that when arguments are missing, functions return "" or a fallback value.
      •	Each test checks for correct fallback behavior (ex, default_output.dat remains unchanged if the option doesn’t exist).
Improvement
      •	Add a test that verifies the error message or the return value when std::stoi fails (GetOptions might return "" or print a message).
      •	Alternate error mechanism: If you plan to move away from printing errors (ex, throwing exceptions or using std::optional), you could add tests using REQUIRE_THROWS or checking for std::nullopt.

Group-06/StringSet/main.cpp
```diff
@@ -0,0 +1,30 @@
...
+    cse::StringSet setIntersection = set1.Intersection(set2);
+    std::cout << "Size of Intersection set: " << setIntersection.size() << "(yellow and blue)" << std::endl;
+
+    set1.clear();
+    std::cout << "check if .clear() make our set empty: " << set1.size() << "(size)" << std::endl;
+    std::cout << set1.empty() << std::endl;
+
+    set2.erase("blue");
+    std::cout << "check if .erase() deleted blue from the set2: " << set2.size() << "(yellow and purple)" << std::endl;
+}
```
caden-james, line 30: Testing and Error Checking - Unit Testing

The tests in main do not cover all functions implemented. I would show some tests for Filter, FilterOut, etc. to make sure those functionalities work.

caden-james: I would also add more edge cases, like Union/Intersection/Difference with empty sets, the filters with no elements provided, etc.

Group-06/StringSet/StringSet.hpp
```diff
@@ -0,0 +1,89 @@
...
+  * @brief Add string into the string set
+  * @param str Added to string set
+  */
+  void insert(const std::string &str) { mElements.insert(str); }
+  
+  /**
+  * @brief Erase specific string from string set
+  * @param str Deleted from string set
+  */
+  void erase(const std::string &str) { mElements.erase(str); }
```
caden-james, line 53: Testing and Error Checking - Assertions

I don't see any assertions here, so you could have an easy assertion here that checks for invalid states like a null string.

caden-james (commented):
Review for StringSet -

Pretty good implementation overall. It's clear how it works and most of my changes are slight tweaks. Implementation is obvious and most improvement could probably just be adding new features to modify the strings.

tdkduong (commented):
Review for ErrorManager
- Coding: The functions are well, there is nothing more I can complain, but I think you can remove some empty lines, since I find it a little bit weird.
- Testing: The tests are good, however, I think there could be an overall test, what I mean is a last test that run through every case, or like when a human try to use your product, and encounter errors.

tranbao01:
Code Review for CommandLine:
1 Design:
1.1 Informative Identifiers
- Clear members and functions' names about what they store and do
1.2 Avoid "Magic Numbers"
- There is no hard coded constants aka magic numbers
1.3 Intuitive and Functional Interface
- The API is clear and straightforward ( addCommand, executeCommand, trackHistory)
- Including a default 'help' command is a nice touch
1.4 Informative Comments
- Solid doxygen-style comments , easy to understand too.
1.5 Consistent Style
- Your brace and indentation style is consistent.
1.6 Visual Appearance
- Code is neatly laid out, with good use of blank lines and indentation.
2. Implementation
2.1 No Warnings During Compilation
- No warnings, nice job
2.2 Proper Uses of Namespaces
- Everything is inside namespace cse. There’s no using namespace std;. 
2.3 Adheres to the C++ Standard
- No undefined behavior is apparent. You are using standard containers (unordered_map, vector) and standard library functionality (sort, function, etc.).
2.4 Appropriate Modularity
- Each method has a single responsibility: adding commands, executing them, printing history, etc.
- The “help” functionality is separated in its own method (printHelp).
2.5 Uses Modern C++ Constructs
- Using std::function is a nice modern approach—very flexible.
You also use range-based logic in printHelp (collecting keys, sorting, etc.). Everything looks current.
2.6 Uses Standard Library Algorithms
- You use std::sort to sort command names in printHelp. 
2.7 Pointer Handling
- Not applicable here; you store everything in standard library containers. No raw pointers are used.
2.8 const Correctness
- Some methods, like printHistory and printHelp, are correctly declared const.
3. Testing and Error Checking
3.1 Unit Testing
- Did cover the main features
3.2 Edge Cases
-The tests do a good job verifying:
+Executing a non-existent command.
+Tracking and printing history when there are commands and when it’s empty.
+State updates in sequential calls.
3.3 Assertions
- The Catch2 macros (REQUIRE, CHECK) are effectively your assertions. They are used correctly to verify expected outcomes or compare output strings.
3.4 Error Handling
- The code in CommandLine::executeCommand handles potential exceptions by printing an error message. 
Overall:
Your code is solid. Your tests are well written that they did cover the main features and edge cases

ellio357:
image

ellio357:
[ErrorManager Class Review.pdf](https://github.com/user-attachments/files/18946123/ErrorManager.Class.Review.pdf)

Group-06/ErrorManager/ErrorManager.cpp
```diff
@@ -0,0 +1,270 @@
...
+  /**
+   * @brief Set the Action that will be invoked when an error of a certain
+   * ErrorType occurs
+   *
+   * @param level Error Level to which action will be assigned
+   * @param action Function that will be called. Does not accept arguments and
+   * returns void
+   */
+  void setAction(ErrorLevel level, const std::function<void()>& action) {
+    assert(action);
```
tdkduong, line 140: I think this line can be removed, since std::function<void()> always contains a callable object.

Group-06/ErrorManager/ErrorManager.cpp
```diff
@@ -0,0 +1,270 @@
...
+   *
+   * @param line Line number where error occured
+   * @param message Message to be sent to user
+   * @param level Error level of a message. Defaults to "Info"
+   */
+  void printError(int32_t line, const std::string& message,
+                  ErrorLevel level = ErrorLevel::Info) {
+    assert(!message.empty());
+    assert(mStringColorCodes.find(level) != mStringColorCodes.end());
+    assert(mTerminationEnabled.find(level) != mTerminationEnabled.end());
```
tdkduong, line 81: assert can be replaced with .at(level), and it will throw std::out_of_range when the key is not existed.

Group-06/ErrorManager/ErrorManager.cpp
```diff
@@ -0,0 +1,270 @@
...
+    if (mLogFile.is_open()) {
+      if (mColorsEnabled) {
+        // Remove ANSI color codes for output in the file
+        size_t pos = 0;
+        while ((pos = output.find("\033[")) != std::string::npos) {
+          size_t endPos = output.find('m', pos);
+          if (endPos != std::string::npos) {
+            output.erase(pos, endPos - pos + 1);  // erase color code
+          }
+        }
```
tdkduong, line 114: I don't remember, but if you are familiar with regex, those line could be replaced. Maybe it could be 
```c++
output = std::regex_replace(output, std::regex("\033\\[[0-9;]+m"), "");
```

Group-06/ErrorManager/ErrorManager.cpp
```diff
@@ -0,0 +1,270 @@
...
+  bool mColorsEnabled = true;
+
+  /// @brief File for logging errors
+  std::ofstream mLogFile;
+
+  /// @brief Map that matches ErrorLevel to the corresponding ANSI color code
+  std::unordered_map<ErrorLevel, std::string> mStringColorCodes = {
+      {ErrorLevel::Info, COLOR_GREEN},
+      {ErrorLevel::Warning, COLOR_YELLOW},
+      {ErrorLevel::Fatal, COLOR_RED}};
```
tdkduong, line 235: I think unordered_map here is unnecessary. std::array can do the same works, since ErrorLevel are fixed, and you can use enum as array index. Such as
```c++
std::array<std::string, 3> mStringColorCodes = {COLOR_GREEN, COLOR_YELLOW, COLOR_RED};

```

tdkduong (commented):
ErrorManager: I have add some comments for the code.

Lilryesx:
Review for CommandLine:


1. Design
- All variable names, function names, and class names are clear and consistent as you use camelCase for your naming convention and do so consistently throughout the class.
- No magic numbers are present as all constants are either strings inputs.
- The code is visually appealing with appropriate spacing and indentation making it easy to follow and read.
- Function names clearly describe their purpose and how they should be used.
- Suggestion:
	Add an overall class-level comment at the beginning of the class to explain the general purpose and intended use of the CommandLine class.

	
2. Implementation
-  The code compiles cleanly with no warnings
- you properly use the namespace cse great job (I forgot to do this lol)
- There are no signs of undefined behavior. Solid adherence to C++ standards.
- Const correctness is applied to functions like printHistory and printHelp

3. Testing and Error Checking
-  overall test suite is perfect! 
- Suggestions:
   add a test that tests for running the same command name twice and ensuring the last one overwrites the previous one.
  add a test that checks for executing commands with empty strings.

Overall: Great job! with a few minor tweaks your code will be closer to perfection!!

matrix373:
The final version of Group06-Lamport classes are implemented. Advanced specs and tests for StaticString and StringSet are added.
Advanced specs are listed in the Specification folder. Advanced tests are listed under tests/Group06-test/ folder named Group06-StaticStringAdvancedTests.cpp and Group06-StringSetAdvancedTests.cpp. Advanced tests are separated from tests for the final version of our classes.

Group-06/StaticString/StaticString.hpp
```diff
@@ -0,0 +1,591 @@
...
+  * @return StaticString& Reference to this StaticString object.
+  */
+  StaticString& operator=(const StaticString& staticString) {
+    // Avoid assigning to itself
+    assert(this != &staticString && "Assigning object itself is not allowed.");
+    if (this == &staticString) {
+      return *this;
+    }
+
+    if (this != &staticString) {
```
mercere99, line 124: This `if` is unneeded. If it were going to be false, the code would have already returned in the previous `if`.  (And if there wasn't a return there, you should just be using `else` anyway).

Group-06/StaticString/StaticString.hpp
```diff
@@ -0,0 +1,591 @@
...
+  *
+  * New string to be assigned must be shorter than size limit or exact length
+  * with maximum size limit. When a StaticString assigned to a nullptr,
+  * it will automatically clear itself and will be empty
+  * as well as empty string initialization.
+  *
+  * @param cstr C-string to be assigned.
+  * @return StaticString& Reference to this StaticString object.
+  */
+  StaticString& operator=(const char* cstr) {
```
mercere99, line 149: Will this work with `std::string` too?  Or will one of the others?  It would be nice to be able to convert an `std::string` to a `StaticString`.

Group-06/StaticString/StaticString.hpp
```diff
@@ -0,0 +1,591 @@
...
+  * @brief Index operator to get value in the specific index of string.
+  *
+  * Index operator only used to get a copy of character on the given index.
+  *
+  * @see set() to change a character on a given index.
+  *
+  * @param index size_t The index of character.
+  * @return Indexed char value.
+  */
+  char operator[](std::size_t index) const {
```
mercere99, line 190: Unless I'm missing something, I think this one can safely be marked `noexcept`.

Group-06/StaticString/StaticString.hpp
```diff
@@ -0,0 +1,591 @@
...
+  }
+
+  /**
+  * @brief Equality comparison with another StaticString.
+  *
+  * @param rhs The StaticString to compare with.
+  * @return true if both strings are equal.
+  * @return false if both strings are not equal.
+  */
+  bool operator==(const StaticString& rhs) const noexcept {
```
mercere99, line 204: Technically you could compare StaticStrings of different maximum sizes since they might be the same current size.  Probably a rare use case though.

Group-06/StaticString/StaticString.hpp
```diff
@@ -0,0 +1,591 @@
...
+  /**
+  * @brief Inequality comparison with another StaticString.
+  *
+  * @param rhs The StaticString to compare with.
+  * @return true if both strings are not equal.
+  * @return false if both strings are equal.
+  */
+  bool operator!=(const StaticString& rhs) const noexcept {
+    return !(*this == rhs);
+  }
```
mercere99, line 231: FYI, I believe that starting with C++20, you don't need to implement `operator!=` if you have `operator==`.

mercere99: ...And if you want to go crazy, look up the spaceship operator for three-way comparisons.  https://en.cppreference.com/w/cpp/language/default_comparisons

It allows you to just implement `operator<=>`, which then handles all 6 of the comparison operators.

Group-06/StaticString/StaticString.hpp
```diff
@@ -0,0 +1,588 @@
...
+    }
+    return std::string_view(mString + start, end - start);
+  }
+
+private:
+  /// @brief Constant value for StaticString null terminator used.
+  static constexpr char null_terminator = '\0';
+
+  /// @brief Array of characters stores the string with a predefined size.
+  char mString[MaxSize + 1] = {null_terminator};  // Array of characters '\0'
```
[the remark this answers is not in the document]
mercere99: The advantage of a raw C-array is that you can pass it out easily, though that might work with an std::array as well (with a free conversion); I'd have to test it.

Group-06/StaticString/StaticString.hpp
```diff
@@ -0,0 +1,591 @@
+/**
+ * @file StaticString.hpp
+ * @author Mehmet Efe Caylan
+ * @brief StaticString Class implementation file.
+ * @version 0.1
+ * @date 2025-02-06
+ *
+ */
```
mercere99, line 8: Overall, this is a very impressive implementation!

Group-06/StringSet/StringSet.hpp
```diff
@@ -0,0 +1,108 @@
+/**
+ * @file StringSet.hpp
+ * @author Orhan Aydin
+ * @brief Header file for StringSet class
+ */
```
mercere99, line 5: Mostly passing functionality through to `std::unordered_set<std::string>`, but a solid job of that, along with a few helpful additions.  My guess is that you're going to need to add a bunch more functions when you start using it in an app (which will become obvious only once you need those functions), but that gives you a reasonable path forward.

mercere99 (approved):
A solid set of classes, nicely done.  Many individual comments inline for things that I recommend you update or fix.

Group-06/ErrorManager/ErrorManager.cpp
IvanBega: Changed unordered map to array

Group-06/ErrorManager/ErrorManager.cpp
IvanBega: Thanks for this suggestion. I have tried to implement regex multiple times, and unfortunately I couldn't make it work properly, even with the help from LLMs. So I decided to leave it as it is

Group-06/ErrorManager/ErrorManager.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+  /**
+   * @brief Executes a function and displays an error if function throws during
+   * execution
+   *
+   * @param function Function to be executed
+   * @param message Error message to be displayed if a function fails
+   * @param level Error level of a message
+   */
+  void executeAndHandleError(const std::function<void()>& function,
```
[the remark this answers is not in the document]
IvanBega: This and printError function overloads call in its body logToFile(), which operates on private member std::ofstream mLogFile. Thus compiler complains if I mark any of these functions as const

Group-06/ErrorManager/ErrorManager.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+  /**
+   * @brief Writes a message to console with a message type at line number
+   *
+   * @param line Line number where error occured
+   * @param message Message to be sent to user
+   * @param level Error level of a message. Defaults to "Info"
+   */
+  void printError(int32_t line, const std::string& message,
+                  ErrorLevel level = ErrorLevel::Info) {
```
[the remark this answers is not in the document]
IvanBega: Added two private helper function to make printError() more readable

tests/Group06-tests/Group06-ErrorManagerTest.cpp
```diff
@@ -0,0 +1,235 @@
...
+#include <iostream>
+#include <sstream>
+#include <stdexcept>
+#include <string>
+
+#include "../../third-party/Catch/single_include/catch2/catch.hpp"
+#define exit(code)          \
+  throw std::runtime_error( \
+      "exit exception")  // redefine exit for testing purposes
+#include "../../Group-06/ErrorManager/ErrorManager.cpp"
```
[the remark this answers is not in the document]
IvanBega: Separated ErrorManager into .hpp and .cpp, and now I use #include .hpp in a test file

PR #64, opened 2025-02-07
Liskov merge

miabodenhorn:
Merging changes from Image Class and Test for Image Class


owenhaiar (approved):
Tested it out, looks good. Formatting doesn't match google style but not a big deal right now

allienello (approved):
Looks good!

PR #77, opened 2025-02-08
[Ritchie] Advanced classes specification + Final working initial classes + Testing

ze-ntsh:


ze-ntsh:
@mercere99 @EvilJuiceBox

ze-ntsh:
@mercere99 @EvilJuiceBox (PR for Initital classes / tests / asserts)

ze-ntsh:
@mercere99 @EvilJuiceBox

tests/Group-02/DataMapTest.cpp
nigamsa1, line 1: Very thorough and good tests.

KrishPatel198:
### **Review for StateGridPosition**

**Design:**

Informative Identifiers: Names are mostly clear. The function create_point could be more descriptive, like construct_point.
Avoids Magic Numbers: The get_object_orientation function has hardcoded 360. Consider using a constexpr variable like constexpr double MAX_ORIENTATION = 360.0.
Intuitive and Functional Interface: Member variables are private, which is good. get_state_grid() should be marked as const. calculate_object_position() should be marked [[nodiscard]].
Informative Comments: Comments are present but should follow Doxygen format for consistency and readability.
Consistent Style: Naming conventions are mostly consistent. create_point is defined inside the class, but others are in the .cpp file—consider moving it for uniformity.
Visual Appearance: Well-structured, good spacing, and indentation. Could benefit from more blank lines to separate logical sections.


**Implementation:**

No Warnings During Compilation: The assertion in calculate_object_position() (assert(m_state_grid == nullptr)) is misleading—should check for nullptr instead.
Proper Uses of Namespaces: No unnecessary using namespace std;, well-contained in cse.
Adheres to C++ Standard: Code follows good practices but could use std::optional<StateGrid*> instead of nullptr for m_state_grid.
Appropriate Modularity: Code is modular, but create_point is redundant—use direct Point initialization instead.
Uses Modern C++ Constructs: std::fmod should replace the while loops in get_object_orientation() for efficiency.
Uses Standard Library Algorithms: No unnecessary loops; std::fmod is a better alternative for angle normalization.
Pointer Handling: m_state_grid is a raw pointer—consider using std::unique_ptr or std::optional<StateGrid*> for safer memory management.
Const Correctness: get_grid_position(), get_object_position(), and get_state_grid() should be const.


**Testing and Error Checking:**

Unit Testing: Ensure unit tests cover movement, orientation normalization, and grid tracking.
Edge Cases: Test negative and large orientation values. Ensure set_object_position(std::pair<int, int>) correctly converts to double without precision loss.
Assertions: Used correctly, but calculate_object_position() should check nullptr before asserting.
Error Handling: Should check for nullptr instead of asserting. Using exceptions or error logging could improve robustness.


**Final Thoughts:**

This class is well-structured and follows good C++ practices. To improve maintainability and efficiency, define constexpr for magic numbers, use std::fmod instead of loops, mark applicable functions as const, replace raw pointers with smart pointers, and use Doxygen-style documentation.

Thisisnotgoingtowork:
## Review for class StateGrid

Ido think these codes are better than mine, really. So I will do what I can, I will try.

### Desing

1. choose_map as a member name can be better named as load_map or something like that, but if you think this is better explaining, you may keep it. you also may what to update the constructor documentation to match the actual parameters used (diff rather than rows and cols). Currently the docstring says “@param rows” and “@param cols” but the constructor only takes a std::string diff. It’s best to keep doc comments in sync with the signature.
2. In choose_map, you might note that if diff is unrecognized, it defaults to "test". It might be helpful if you can quick comment that for the reader.

### Implementation

1. If you plan to support multiple map “difficulties,” you might eventually move them out to separate data files or a resource manager. Right now, it’s all in a hard-coded std::map<std::string, std::vector<std::string>>

### Testing & Error Checking

1. Good testing, you may want to add some tests like for an unrecognized difficulty string (ensuring it defaults to “test”) in the future.


That all I have now, overall, I think you did fulfill most of the requirements, those comments are just some suggestions only.

Group-02/include/cse/StateGrid.h
```diff
@@ -0,0 +1,49 @@
...
+  ///Grid to represent game map
+  std::vector<std::string> m_grid;
+  ///Map to represent the name and (future) properties of each state <<< Soon to be DataMap type
+  std::map<char, std::vector<std::string>> m_dictionary = {
+      {' ', {"EmptySpace", "Open"} },
+      {'#', {"Wall", "Closed"}},
+      {'X', {"Enemy", "Open"}},
+      {'0', {"Exit", "Open"}},
+      {'P', {"Player", "Closed"}}
+  };
```
gringrant, line 30: Instead of a vector of strings that holds 2 properties, I would recommend making a dedicated struct that holds all the properties each character needs.

chrispaul-lab:
**Review for DataMap** 

_**Design**_

_Informative identifiers:_
1. The DataMap class uses clear and descriptive identifiers for variables and functions.
2. Names effectively communicate the purpose of each component, making the code readable.

_Avoids “magic numbers”:_
1. No arbitrary values are present that would benefit from named constants:
2. No hardcoded numbers

_Functional interface:_
1. The interface follows standard container patterns
2. Common operations like insertion, retrieval, and existence checking are available.
3. Member variables are properly encapsulated

_Informative comments_
1. Documentation is informative.
2. Includes file headers, function documentation, and namespace.

_Consistent style_
1. The code consistently uses the m_ prefix for member variables.
2. Function names follow camelCase convention.
3. Parameter naming remains uniform.

_Visual appearance_
1. Well-structured
2. It is visually easy to understand.


_**Implentation**_

_No warnings during compilation_
1. Everything compiled with no warnings & no compiler issues. 

_Namespace use_
1. All code is enclosed within namespace cse { ... }.
2. Avoids using namespace std, preventing unintended namespace pollution.
3. Ensures that header files do not introduce unwanted global scope contamination.

_C++ standard_
1. Follows modern C++ best practices.
2. No obvious instances of undefined behavior
3. Uses standard library components correctly

_Appropriate Modularity_
1. Clearly separates concerns and responsibilities within the class.
2. Methods are concise and well-structured.
3. Some methods are implemented inline in the header file, which is appropriate for templates.
4. Could benefit from additional helper functions to encapsulate repetitive logic.

_Uses modern C++ constructs_
1. Utilizes std::any for type erasure.
2. Uses range-based for loops where applicable.
3. Applies modern attributes like [[nodiscard]] and [[maybe_unused]] when relevant.
4. Supports std::initializer_list for ease of initialization.


_Uses standard library algorithms_
1. The implementation relies on direct container operations rather than STL algorithms.
2. Could improve readability and efficiency by including std::find_if, std::transform, std::any_of, etc.
3. Manual iteration could be replaced with more expressive standard library functions.

_Pointer handling_
1. Avoids manual memory management with raw pointers.
2. Uses standard containers that handle memory automatically.
3. Follows value semantics properly, using std::any where needed.

_const correctness_
1. Properly marks member functions as const when they do not modify the state.
2. Uses const& for parameters where modification is unnecessary.
3. Maintains const correctness throughout the class.
4. operator[] remains non-const as expected since it modifies the container.

_**Testing & Error Checking**_

_Unit testing_:
1. Coverage is strong but could explicitly test all public member functions, including constructors.

_Edge cases_
1. Covers key cases (empty map, non-existent keys, type verification) but could expand boundary condition testing.

_Assertions_
1. get method should enforce type checking with assertions, as indicated by the test file but missing in implementation.

_Error handling_
1. at should throw an exception for missing keys, like std::map::at.
2. get should handle type mismatches with assertions or exceptions.

Group-02/include/cse/StateGrid.h
```diff
@@ -0,0 +1,49 @@
...
+
+  char get_state(int row, int col);
+
+  std::vector<std::string> define_state(char state);
+
+  bool validate_position(std::pair<int, int> move);
+
+  std::vector<std::pair<int,int>> find_moves(int row, int col);
+};
+}  // namespace cse
```
gringrant, line 49: I see you use a lot of std::pair<int, int>'s I would recommend having a point struct with an x and y int as members. It would make the code much easier to read.

Group-02/src/cse/StateGrid.cpp
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * @brief Displays the grid to user
+ */
+void StateGrid::display_grid() {
+  assert(!m_grid.empty() && "m_grid is empty and cannot display");
+
+  for (const std::string& line : m_grid) {
+    std::cout << line << "\n";
+  }
+}
```
gringrant, line 32: Idea: overload the << operator for an out stream so you can just `cout << myStateGrid;`

Group-02/src/cse/StateGrid.cpp
```diff
@@ -0,0 +1,123 @@
...
+bool StateGrid::set_state(std::pair<int, int> new_position, std::pair<int, int> agent) {
+  assert(!m_grid.empty() && m_grid[agent.first][agent.second] == 'P');
+
+  if (validate_position({new_position.first, new_position.second})) {
+    m_grid[new_position.first][new_position.second] = 'P';
+    m_grid[agent.first][agent.second] = ' ';
+    return true;
+  }
+  return false;
+}
```
gringrant, line 48: Instead of have a bunch of .first and .second I would recommend making a point struct with x and y members.

Group-02/src/cse/StateGrid.cpp
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Returns property information about a specific state
+ * @param state queried state
+ * @return vector of strings holding info about state
+ */
+std::vector<std::string> StateGrid::define_state(char state) {
+  auto possible_state = m_dictionary.find(state);
+  assert(possible_state != m_dictionary.end() && "This state is not in the map!");
+  return m_dictionary.at(state);
+}
```
gringrant, line 59: Would recommend using an std::option to represent the program state where no StateGrid state was found.

Group-02/src/cse/StateGrid.cpp
```diff
@@ -0,0 +1,123 @@
...
+bool StateGrid::validate_position(std::pair<int, int> move) {
+  assert(move.first < m_rows && move.second < m_cols && "This move is out of bounds");
+  char validate = m_grid[move.first][move.second];
+  if (m_dictionary.find(validate) != m_dictionary.end() && m_dictionary.at(validate)[1]=="Open")
+  {
+    return true;
+  }
+
+  return false;
+}
```
gringrant, line 86: Instead of having if(_expression_) {return true} else {return false}
I would recommend simply using return _expression_
Much easier to read.

Group-02/include/cse/AuditedVector.h
```diff
@@ -0,0 +1,67 @@
...
+// In release mode, just use std::vector.
+template <typename T, typename Allocator = std::allocator<T>>
+using AuditedVector = std::vector<T, Allocator>;
+
+#else
+template <typename T, typename Allocator = std::allocator<T>>
+class AuditedVector : public std::vector<T, Allocator> {
+ public:
+  using base_t = std::vector<T, Allocator>;
+  using base_t::base_t;
```
132ikl, line 25: The second `using base_t::base_t` line is a bit surprising (but does seem to be necessary), a small explanation comment might be helpful.

The usage of a `using` declaration to the base type makes the class very clean :smile:

Group-02/include/cse/AuditedVector.h
```diff
@@ -0,0 +1,67 @@
+/**
+ * @file AuditedVector.h
+ * @author Matthew Hawkins
+ */
+
+#ifndef AUDITEDVECTOR_H
+#define AUDITEDVECTOR_H
```
132ikl, line 7: Consider using `#pragma once` like the rest of your team's classes

Group-02/include/cse/AuditedVector.h
```diff
@@ -0,0 +1,67 @@
...
+
+  T& operator[](size_t index) {
+    assert(index < this->size() && "cse::vector: index out of range");
+    return base_t::operator[](index);
+  }
+
+  const T& operator[](size_t index) const {
+    assert(index < this->size() && "cse::vector: index out of range (const)");
+    return base_t::operator[](index);
+  }
```
132ikl, line 35: It might be helpful if you also printed how many items are actually in the vector, like `std::vector::at` does

tests/Group-02/AuditedVectorDebugTest.cpp
```diff
@@ -0,0 +1,42 @@
+/**
+* @file AuditedVectorTestDebug.cpp
+ * @author Matthew Hawkins
+ *
+ * Note: In debug mode the audited vector uses assert to check preconditions.
+ * Because an assertion failure aborts the program, we only test valid operations here.
```
132ikl, line 6: You could consider using my group's [assert macros](https://github.com/CSE498/CSE498-Spring2025/blob/group-4/Group-04/lib/CseAssert.hpp) directly, or using a similar technique as [our macro tests](https://github.com/CSE498/CSE498-Spring2025/blob/group-4/tests/Group-04/Assert.cpp#L3-L9) to test for assertions being properly triggered without aborting the program.

tests/Group-02/AuditedVectorReleaseTest.cpp
```diff
@@ -0,0 +1,49 @@
+/**
+* @file AuditedVectorTestRelease.cpp
+ * @author Matthew Hawkins
+ *
+ * In release mode, our audited vector is just an alias for std::vector.
+ * Thus, we test its behavior as a normal vector.
+ */
+
+#define NDEBUG
+#include <cassert>
```
132ikl, line 10: This include shouldn't be necessary

tests/Group-02/AuditedVectorReleaseTest.cpp
132ikl, line 1: Consider adding a test case which requires, e.g., `std::is_same_v<cse::AuditedVector<int>, std::vector<int>>` to be true in release mode

132ikl (commented):
I reviewed `AuditedVector`, overall it looks great! Functionality seems to be in place and it compiles with no warnings. Well done :smile:

CircleShift (commented):
Design:
- Informative identifiers: m_actions is informative, and all functions have reasonable argument names.
- Magic Numbers: None seen in code.
- Interface: Simple and easy to understand function names.
- Comments: All comments written in doxygen style, and are really informative.  Well done!
- Consistent style: Using clang-format.
- Visual appearance: Code is easy to parse.

Implementation:
- Warnings during comp: No warnings thrown by ActionMap.h, nor by ActionMap.cpp
- Proper uses of namespaces: Everything is in the cse namespace.
- Adheres to c++ standard: Majority of unordered_map does not produce UB, and those that do are guarded (see invoke)
- Appropriate modularity: Code is quite simple and usable as is.
- Uses modern c++ constructs: using std::any and std::function, lambdas in tests
- Uses standard library algorithms: None used, but functions are simple enough to not require them.
- Pointer handling: No pointers directly defined in code
- const correctness: Everything marked const makes sense.  You may want to see if you can make invoke const since it does not alter the inner map by default (though I can see the reasoning to not do so, since an invoked function may reference the ActionMap itself).

Testing:
- Unit testing: Functions are comprehensively tested.  You may want to try some invalid things so you can see how the class behaves under those circumstances.
- Edge cases: Some things you may want to test for (invoking actions that do not exist, erasing actions that do not exist).
- Assertions: There is an assertion around the any cast in invoke to make sure the function signature is correct.
- Error handling: Asserts used for invoke, other map functions do not produce errors.

garciaow:
Code Review for StateGridPosition
Owen Garcia

# Design
**1. Informative identifiers:**
Nearly everything has a name that works with a consistent style. I would suggest making the Point struct's x and y a tad bit more descriptive; maybe x_position and y_position (or even add a point_ in front)? m_central_position could also be made into something like m_grid_central_position.

**2. Avoids "magic numbers":**
I did notice a 360 in the .cpp file. This could easily just be made into a constexpr variable.

**3. Intuitive and functional interface:**
I think everything looks good here. The getters could maybe be made const?

**4. Informative comments:**
Comments are informative. I can't quite think of anything else to add here.

**5. Consistent style:**
Consistent use of snake_case. Everything looks fine here.

**6. Visual appearance:**
Code is readable and well-structured.

# Implementation
**1. No warnings during compilation:**
I had no warnings when I ran this.

**2. Proper uses of namespaces:**
Everything is contained in namespace cse and there is no use of namespace std. Good work.

**3. Adheres to the C++ standard:**
The code adheres to the standard. All good here.

**4. Appropriate Modularity:**
I'm not too sure create_point is needed; couldn't Point just be initialized directly?

**5. Uses modern C++ constructs:**
Consider std::optional<Point> instead of Point(0, 0) default values. Also, instead of raw pointer StateGrid* m_state_grid, you could try std::unique_ptr<StateGrid> or std::shared_ptr<StateGrid>.

**6. Uses standard library algorithms:**
std::fmod could be used in get_object_orientation() instead of while loops.

**7. Pointer handling:**
m_state_grid is a raw pointer; if there are no ownership semantics worked out in future implementations, I'd suggest using a unique pointer.

**8. Const correctness:**
Getters could be made const.

# Testing and Error Checking
**1. Unit testing:**
I'd suggest adding a test for the set_object_position <unit, unit> overload.

**2. Edge cases:**
I see the cases for values outside of the 0-359 range. Might be redundant, but maybe you'd consider a check that 360 returns 0.

**3. Assertions:**
I'd suggest checking for the null pointer before asserting in calculate_object_position.

**4. Error handling:**
I'd suggest checking the StateGrid* m_state_grid null pointer before use.

**Other Notes**
I'm not sure how this would work in the context of your project, but I'd suggest discussing implementing grid boundaries with your team. Could an absurdly high value lead to any issues down the line?

iamrb1:
@mercere99 @EvilJuiceBox advanced specification and the final version of the initial classes are here for grading purposes

Group-02/src/cse/DataMap.cpp
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Operator overload for Datamap []
+ * defaults to 0 if called DataMap["key"]
+ * Allows for assignment DataMap["key"] = 90
+ * @param key
+ * @return std::any&
+ */
+std::any& DataMap::operator[](const std::string& key) {
+  if (!m_map.contains(key)) {
+    m_map[key] = std::any(0);
```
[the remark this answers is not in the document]
mercere99: Zero can be thought of as a valid default; it doesn't necessarily need to be changed, but changing it might make it clearer.  For example, when working with pointers, nullptr is basically always a 0 (though it technically doesn't have to be), but they originally started calling it NULL to make functionality clear.  In this case you might want something like `static constexpr int DEFAULT_VALUE = 0;` and then use DEFAULT_VALUE there.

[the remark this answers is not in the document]
mercere99: Zero can be thought of as a valid default; it doesn't necessarily need to be changed, but changing it might make it clearer.  For example, when working with pointers, nullptr is basically always a 0 (though it technically doesn't have to be), but they originally started calling it NULL to make functionality clear.  In this case you might want something like `static constexpr int DEFAULT_VALUE = 0;` and then use DEFAULT_VALUE there.

Group-02/include/cse/ActionMap.h
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * @brief Invoke an action with arguments
+   * @tparam Ret Return type
+   * @tparam Args Argument types
+   * @param action Action name
+   * @param args Arguments
+   * @return Return value of the function
+   */
+  template <typename Ret = void, typename... Args>
+  Ret invoke(const std::string& action, Args&&... args) {
```
mercere99, line 54: @CircleShift points out that this function can probably be made `const`.  Not sure if you tried and had problems with that, but I'm happy to help if there's something I'm missing right now that makes it trickier than I expected.

Group-02/src/cse/StateGridPosition.cpp
```diff
@@ -0,0 +1,69 @@
...
+ * @author Bram Hogg
+ */
+
+#include <cse/StateGridPosition.h>
+#include <cassert>
+#include <cmath>
+
+namespace cse {
+
+const int MAX_ROTATION = 360;
```
mercere99, line 12: You should make this variable `constexpr` so that this value is propagated at compiler time rather than run time (will learn about this technique properly in a couple of weeks).

mercere99 (approved):
Overall, this is a reasonable set of classes.  Your code shouldn't interfere with other groups, so I'm going to go ahead and merge it into main.

Group-02/include/cse/DataMap.h
```diff
@@ -0,0 +1,161 @@
...
+   * Get value from key std::string
+   * If key does not exist, set key to defaulted any templated value constructor
+   * @tparam T templated value returned
+   * @param name std::string
+   * @return T
+   */
+  template <typename T>
+  [[maybe_unused]] T get(const std::string& name) {
+    if (!m_map.contains(name)) {
+      m_map[name] = std::any(T());
```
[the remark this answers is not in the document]
iamrb1: Change added to DataMap.h!

Group-02/include/cse/DataMap.h
```diff
@@ -0,0 +1,161 @@
...
+  /// Member Variable unordered_map
+  std::unordered_map<std::string, std::any> m_map;
+
+ public:
+  /**
+   * Default constructor
+   */
+  DataMap() = default;
+
+  [[maybe_unused]] DataMap(std::initializer_list<std::pair<std::string, std::any>> initial);
```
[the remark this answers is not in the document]
iamrb1: Change added to all relevant Methods in DataMap. For the constructor, however; I was unable to add inline or static as marked within this comment.

Group-02/include/cse/DataMap.h
```diff
@@ -0,0 +1,161 @@
...
+  /**
+  * .at method for finding where key is
+  * @tparam T templated value returned
+  * @param name std::string
+  * @return
+  */
+  template <typename T>
+  [[maybe_unused]] T at(const std::string& name) {
+    assert(contains(name) && "Key does not exist in DataMap");
+    const std::any& val = m_map[name];
```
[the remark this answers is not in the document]
iamrb1: Change added and .at is now const!

Group-02/include/cse/DataMap.h
```diff
@@ -0,0 +1,161 @@
...
+    assert(val.type() == typeid(T) && "Wrong type requested from what is contained within DataMap for value");
+    return std::any_cast<T>(m_map[name]);
```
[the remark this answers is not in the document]
iamrb1: Should have done this in the first place thank you for catching this! Change added.

Group-02/src/cse/DataMap.cpp
```diff
@@ -0,0 +1,46 @@
...
+* @param name std::string name to be found
+* @return bool if key exists or not
+ *
+ * (For grading purposes of initial specs, this function function was renamed from keyContains to better fit naming
+ * convention)
+*/
+bool DataMap::contains(const std::string& name) const {
+  return (m_map.find(name) != m_map.end());
+}
+
```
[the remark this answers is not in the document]
iamrb1: Moved contains to header file as suggested!

Group-02/src/cse/StateGrid.cpp
DominikLeisinger: Fixed, thank you!

Group-02/src/cse/StateGrid.cpp
```diff
@@ -0,0 +1,119 @@
...
+ * @return vector of strings holding info about state
+ */
+std::vector<std::string> StateGrid::define_state(char state) {
+  auto possible_state = m_dictionary.find(state);
+  assert(possible_state != m_dictionary.end() && "This state is not in the map!");
+  return m_dictionary.at(state);
+}
+///REVIEW COMMENT: A review comment was made to add std::optional for this function in the case where no
+/// StateGrid state was found, but I do not believe this necessary as the assert checks
+/// if the row,col is within bounds which will guarantee a StateGrid state.
```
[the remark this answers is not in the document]
DominikLeisinger: Other than for testing, this function would only be called with valid row/col combinations, so a call that is outside of the grid should "never" happen and I would want the application to crash in this instance.

Group-02/src/cse/StateGrid.cpp
```diff
@@ -0,0 +1,119 @@
...
+namespace cse {
+
+/**
+ * @brief Contructor for StateGrid object
+ * @param diff string representing requested difficulty
+ */
+StateGrid::StateGrid(const std::string& diff) {
+  load_map(diff);
+}
+
```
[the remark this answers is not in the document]
DominikLeisinger: Done, thank you!

PR #79, opened 2025-02-09
Liskov

owenhaiar:
Group-09 changes from last week.

Initial code implementations and some basic unit testing


owenhaiar:
@EvilJuiceBox

Group-09/Image/image.hpp
```diff
@@ -0,0 +1,58 @@
+/**
+* @file image.hpp
```
LeoSSpecht, line 2: Following the pattern from the rest of the files I would rename the file to `Image.cpp` and `Image.hpp`

Group-09/Image/image.hpp
```diff
@@ -0,0 +1,58 @@
...
+ * @brief A struct to represent an image
+ */
+class Image {
+private:
+    std::string url;
+    int width;
+    int height;
+    std::string altText;
+
+    // bool validateURL(const std::string& url) const;
```
LeoSSpecht, line 23: Also recommend removing unused code to keep the overall structure cleaner!

Group-09/Image/image.hpp
```diff
@@ -0,0 +1,58 @@
...
+  /**
+  * @brief Compares two Image objects and determine if they are the same values
+  * @param image to compare
+  * @return true if same values
+  */
+    bool operator==(const Image &image) const {
+       return ((image.url == url) && (image.width == width) && (image.height == height));
+    }
+
+  /** @return The url of the image */
```
LeoSSpecht, line 37: The indentation here is a bit off. This is very slight.

I suggest some auto-formatter to fix this automatically!

LeoSSpecht:
# Review of `Image`

## TLDR
The code is well-structured with clear naming and modular design, with minor formatting issues to be improved upon.

The interface is intuitive but could have additional functionality and more tests, especially for resizing and outputting. 

Removing commented-out code would improve readability, and integrating tests into the Makefile would streamline development.

## Design

### Informative identifiers
In general I think naming conventions are clear and concise. They are also very consistent throughout the files.

The only naming issue I saw was that the file was named `image.hpp`, while other files were capitalized.

### Avoids “magic numbers”
There was no to small use of constants in general, so there were no problems related to having magic numbers throughout the code.

### Intuitive and functional interface
In general the interface seems ok. I would live to have seen a bit more use on the test cases, or if visual use is better maybe some screenshots on how EMScripten renders the images.

One suggestion that that instead of returning a string the user can use a function where they pass in a `std::stringstream` and it writes the JS code or the HTML into that stream, making some functions more intuitive.
I would also recommend some way of adding indentation. When coding the spacing is very important, but adding indentation to the current way of exporting would be a bit troublesome.

A suggested interface could be:
```cpp
int main(){
  Image im(url, 10, 10);
  
  im
    .setCodeIndentation(4)
    .setElementId("my_id"); // Setting the element id for HTML or JS
  im.toHTML(std::cout); // Prints correct if and with correct indentation
}
```

Another useful interface that could be more complete is when resizing the image. This is a very complex feature as people want different things to happen when resizing images (should it crop, should it keep the aspect ratio and not follow exact size, is so which dimension should it follow height or width, etc.)
### Informative comments
The docstrings are very good and consistent.

There aren't many comments, but I don't think they are needed. One part that lacks comments that could be used is on the resize function, where different things can be done to the size of the image.

### Consistent style
Not much to say here. Spacing, naming and formatting is consistent in general. I left 1 comment about the spacing in one line, but it was a small thing.

### Visual appearance
Overall the functions are concise and handle only one thing, which is great.
The only thing I would change is to remove bits of commented code. They make the code look a bit cluttered in general.

## Implementation

### No warnings during compilation
When running the tests the Image class tests were not included originally on the Makefile so I was not able to check for any warnings initially.

However I tried to change the Makefile and remove the `emscripten` imports and was able to successfully run the tests.

I recommend to fix this and include the tests on the Makefile to make the whole process easier on the reviewers!

### Proper uses of namespaces
Code is inside `namespace cse`. There is no namespace pollution and importing of the whole std library.

### Adheres to the C++ standard
I could not find any instances of undefined behavior or anything that is breaking standards or making the code unsafe. As it does not have to handle pointer it does not have any possible memory leaks or related issues.

### Appropriate Modularity
In general the code is modular and divided into well defined functions. The only places I would recommend to avoid duplicating code is on the constructor. The constructor does the same validation as `setURL`, so `setURL` could be used instead, or moving the validation logic to a specific function.

### Uses modern C++ constructs
There is not a big use of new features , however there was also no big need for them in general.
It does use some emscripten features, that could be considered new.

### Uses standard library algorithms
In general I don't think the implementation had much space for using the standard algorithms.

### Pointer handling
There is no use of pointers in this class.

### const correctness
All getters are const correct, and where applicable const and references are used.

## Tests and error checking

### Unit testing.
Tests are mostly lacking in my opinion. There are only 4 asserts, which test the getters. However the tests were not being ran on the Makefile.

I recommend adding tester for the setters, testing the invalid inputs, testing the outputs like HTML and JS and also testing the resize functionality, specially because the resize can get quite complex.

### Edge cases.
As memory is not handled there are less possible edge cases. There are also not a lot of logical edge cases to be checked for.

### Assertions Error handling
There are no asserts. But in some places error handling was valid and correct in my opinion.
I am not sure there would be a lot of use for asserts in this case.

typritchett6:
**### Review of Group9 - WebLayout** 

**Overall**
The WebLayout class is very solid overall and does a good job of managing/rendering elements correctly. Some minor improvements could be made to enhance performance and readability. 

**Design**
•	Informative identifiers – The functions PushTextBox() and PushImage() could be named something a little more informative like RenderTextBox() and RenderImage() since push could mean different things. Could also rename the variables x and y to xPos and yPos or something similar to explicitly say they refer to positions. 
•	Avoids “magic numbers” – The JavaScript portion of code contains a few magic numbers (I’m not sure if this is the same since it’s in javascript within C++) and you could consider representing them at the top of the file as const variables for the max width and heights. That would also make it easier to update them if needed in the future.
•	Intuitive and functional interface – It looks as if the class is easy to add and remove images and text boxes while also including private member variables. You could also add a function to potentially clear the layout if that is something that should be possible. It looks like the TransitionToPage() function is also not implemented but is declared. 
•	Informative comments – Probably don’t need comments next to the asserts in the PushTextBox() and PushImage() functions. The structs in your header file could include a comment at the top explaining what they’re used for. 
•	Consistent style – TransitionToPage() and LoadPage() are the only functions that start with a capital letter so those should probably be made lowercase for consistency (use camelCase for all function names).
•	Visual appearance – Looks great visually.


**Implementation**
•	No warnings during compilation – No big warnings. Just when running code with -Werror you get:
./WebLayout.h:21:32: error: field 'x' will be initialized after field 'y' [-Werror,-Wreorder-ctor]
   21 |   TextBoxLayout() : textBox(), x(0), y(0) {};
    |                                ^~~~  ~~~~
      |                                y(0)  x(0)

which is basically just saying since you declare y before x in the image and textbox layout structs, you should initialize them in that order as well. 
•	Proper uses of namespaces – Code properly handles namespaces (all code within namespace cse{} and avoids generic using namespace std and other namespaces in header files.
•	Adheres to the C++ standard – might consider making your asserts in PushTextBox() and PushImage() have the values be >= 0 since isn’t (0,0) a valid position normally?
•	Appropriate Modularity – Looks good. Could have a templated helper function to replace the similar loops in LoadPage(). Something like:

template<typename LayoutType> void RenderElements(const std::vector<LayoutType>& elements, void (*renderFunc)(const LayoutType&)) {
 for (const auto& element : elements) { 
if ((element.image.getHeight() > 0 && element.image.getWidth() > 0) && (element.x >= 0 && element.y >= 0)) { 
renderFunc(element); 
} 
} 
}
•	Uses modern C++ constructs – Can use auto when looping using iterators:  
for (const auto& textBox : textboxes) instead of using iterater it = textBoxes.begin() …
•	Uses standard library algorithms – Looks good!
•	Pointer handling – use of raw pointers in main.cpp could be avoided to avoid memory leaks (change to smart  (unique) pointers).
•	Const correctness – getImages() and getTextBoxes() are currently returning copies of their vectors, this could be changed to return (const &) references to avoid unnecessary copies.


**Tests and Error Checking**
•	Unit testing – Unit tests don’t seem to be present in the LoadPage() function. Would be smart to add some testing for that since it’s a public member function. 
•	Edge cases – Could add tests for if Layouts are added with width/heigh of 0 and if a text box has an empty string. Could also test if an image is added twice, what should happen and does removing the image remove both instances or just one?
•	Assertions – Asserts are correctly and effectively used throughout the code. (see adheres to C++ standards above).
•	Error handling – removeImage() and removeTextBox() should return a bool to indicate if it was successful or not. Should ImageLayout or TextBoxLayout’s be checked if they have invalid (negative for example) dimensions when they are added?

Muelleralex1021:
### Review for EventQueue
**Design**
Informative identifiers
all names of variables, functions and types are clear and have a consistent format. everything here looks good.

Avoids “magic numbers”
I did not see the use of any magic numbers so this is good.

Intuitive and functional interface
Class is very straight-forward and only allows for necessary functionality.
Function names are clear to how they are intended to be used.
Member variables are private. 

Informative comments
Comments from function to function are all informative.
maybe add comments to the two private member variables for a little more clarity. 

Consistent style
The style is very consistent throughout this class.

Visual appearance
header file is a little difficult to read, maybe consider adding new lines between functions. 

**Implementation**
No warnings during compilation
Testing looks great!

namespaces are good.

Adheres to the C++ standard
All functions seem to handle exceptions well and limit undefined behavior.

Appropriate Modularity
looks good. not really necessary as functions are all very short.

Uses modern C++ constructs
Good use of modern C++ practices

Uses standard library algorithms
all good.

Pointer Handling
No pointers in use, so this one is good as well.

Const Correctness
Const correctness seems to be consistent across the class.

Unit testing
tests are all present for each function and are all very detailed. 

Edge cases
edge cases are all covered for.

Assertions
Didn't see any assert usage. 

Error handling
error handling is done well and is clear to the user.

ephraimbennett:
# WebLayout
## Design
### Informative identifiers. 
Names of variables and functions are clear enough to understand all. You may want to consider a naming convention for member variables for clarity however. 
### Avoids “magic numbers”. 
There are some magic numbers inside the javascript calls. Consider replacing with constants defined at the top of the cpp file or maybe a constant within the javascript? All the other magic numbers outside the javascript are just checking if a value is positive, which I don't believe counts.
### Intuitive and functional interface. 
The class exposes only the necessary functionality such as adding and removing textboxes, getters for the vectors, and loading pages. Function names are clear enough and good usage of overloading for the structs.
### Informative comments.
Great use of doxygen to describe functions, comments provide perfect clarity and are not overly reduntant.
### Consistent style. 
Local variables and member variables use camelCase. Functions are **not** consistent. PushTextBox, PushImage, LoadPage, and TransitionToPage use PascalCase, whereas addImage, removeImage, etc. are all camelCase. Consider picking one and sticking with it.
### Visual appearance. 
The code is easy to scan through and visually well structured. There seems to be consistent limits on line width.

## Implementation
### No warnings during compilation.
warning: field 'x' will be initialized after field 'y' [-Wreorder-ctor]
   38 |   ImageLayout() : image("", 0, 0), x(0), y(0) {};
      |                                    ^~~~  ~~~~
      |                                    y(0)  x(0)
There are many warnings just like this. Quick fix is to just change the order x and y are declared in WebLayout.h
### Proper uses of namespaces. 
All code is within namespace cse. No usage of "using namespace std;"
### Adheres to the C++ standard. 
The code avoids undefined behavior and runs on differnent compilers and platforms.
### Appropriate Modularity. 
Common procedures are grouped together for the most part, but you could save code duplication from the TextBoxLayout and ImageLayout by implementing a superclass. It might require a bit of work but something worth looking into. 
### Uses modern C++ constructs. 
The code uses modern c++ constructs where applicable.
### Uses standard library algorithms.
There doesn't seem to be any extraneous loops where the standard library functions could fill in.
### Pointer handling. 
No usage of dynamic memory or pointers that need handling, so it's all good.
### const correctness. Are member functions to const if they don’t change the state of the class? Are parameters (especially reference parameters) marked const if they don’t change inside the function?
PushTextBox and PushImage should be marked as const. 
## Testing and Error Checking
### Unit testing. 
Unit tests are present for all public member functions except for LoadPage, which should be added. The other test cases will exercise private portions of the code and are a good test.
### Edge cases. 
Does testing cover most edge cases, as possible? For example, for pointers, does it test null pointers (if they are allowed)? Does it test for empty containers being used? Does it test for 0 or other boundary conditions on parameters?
Negative values should be included in the test cases for things like image sizes and such. 
### Assertions. 
Assertions are properly used throughout the test cases.
### Error handling. 
Errors are properly checked for.

Devon-FoxElster:
# Review for TextBox (And FormattedText)

## Design
### Informative identifiers
All identifiers look good.
### Avoids “magic numbers”
TextBox: All good
FormattedText: There were some magic numbers in the getEscapedText() function relating to the lengths of the escape characters. This could be mitigated by storing them as static constexprs at the top and using them as normal in the function.
### Intuitive and functional interface
TextBox: Overall the interface works, but I was a little confused by the use of the TextBoxConfig in the constructor. I like the idea of a Config object, but it looks like it gets deconstructed anyway right in the constructor. Maybe you could store a Config within textBox to preserve all of the settings? You could also feature a constructor with the 3 inputs as separate parameters. I think a combination of these two ideas would give the most flexibility

FormattedText: The interface for FormattedText is a little confusing, as there are 3 separate getter functions for the text within the class. It isn't specified which functions should be in use for different use cases, but it seems like getEscapedText is the only function to actually change the string before returning it. There may be an opening to simplify this to just getText(), where if it is of RAW type it will escape the characters before returning?
### Informative comments
All of the inline comments are good, although some of the Doxygen comments could use some expansion around the parameters and returns beyond what type they are. In particular, informing the user that the TextBox(.resize() function defaults to 0 for negative integers would be helpful information.
### Consistent style
All variable, class, and function names seem to be consistent among themselves, but I did notice that TextBox uses #ifndef while FormattedText uses #pragma once. I recommend choosing one and standardizing across the files.
### Visual appearance
All of the functions are well-structured and easy to read
## Implementation
### No warnings during compilation
Looks good!
### Proper uses of namespaces
 Everything looks to be contained within the cse namespace
### Adheres to the C++ standard
Looks good to me
### Appropriate Modularity
This was addressed in the intuitive interface section, but the method FormattedText uses to give the user the text could be refined
### Uses modern C++ constructs
Looks good
### Uses standard library algorithms
std::replace_if() could be helpful within the getEscapedText() function as a way of replacing characters with their escaped versions
### Pointer Handling
Looks good
### Const Correctness
Looks good

## Testing and Error Checking
### Unit testing
Test cases do a good job of covering all functions, one area they could be improved would be testing error conditions more often. Especially with escaped characters, testing the output of getText() vs. GetEscapedText() on the same string could be a useful benchmark to compare the two.
### Edge cases
Not many edge cases that I can think of that aren't covered in the tests.
### Assertions
Not many assertions within the classes, but admittedly I didn't see many opportunities for them. One place that has potential is TextBox's resize() function, you could assert that the size is greater than the minimum allowed before switching them. You could also add an assertion to clearText() to make sure it defaults to an empty string.
### Error handling
Looks good!

### Other suggestions
Based on the description of TextBox on the class website it should be able to handle formatted text, and from the looks of it you branched that section into its own class for good modularity. From here I would recommend integrating the two classes together by adding a FormattedText object to TextBox to give it the needed functionality. You could pass the user's string directly into the FormattedText constructor as well during the initial call.
Another small fix would be for the TextBox default constructor. I would recommend removing the variable sets from the constructor and just assigning them to the private variables, i.e.
```
  TextBox() = default;

private:
  std::string text = "";
  int width = 0;
  int height = 0;
```

AmeryJohnson:
# Liskov - Image Review
**Design**
**Informative Identifiers:** Yes informative and understandable.
**Avoid Magic Numbers:** Yes
**Intuitive and functional interface:** Mostly but since it uses emscripten it is not completely intuitive since you need to have emscripten installed and running properly in order for this class to work, although I don't know another way to make this class work without it.
**Informative Comments:** Comments are informative but I think there are some functions where you could explain a bit more for example the functions that only have a return comment on what it returns. 
**Consistent Style:** Style is consistent yes.
**Visual appearance:** Good appearance and proper spacing.

**Implementation**
**No warnings during compilation:** No warnings but the emscripten import is an issue if people do not have it configured correctly and the makefile did not have the test files. 
**Proper uses of namespaces:** Yes
**Adheres to the C++ Standard:** No undefined behavior that popped out to me
**Appropriate Modularity:** Yes all functions are defined clearly and used.
**Uses C++ modern constructs:** I don't think it needs to other than emscripten.
**Uses for standard library algorithms:** There might be a rare case where an STL algorithm could be used but everything looks like it didn't need any.
**Pointers:** None
**Const correctness:** Everything marked const seems to be correct.

**Tests and error checking:** 
**Unit testing:** Very minimal unit tests. I would suggest adding many more tests and for every function as well and test for invalid inputs etc... there is a lot of room for more tests in my opinion.
**Edge Cases:** I can't think of any edge cases associated with images other than maybe like 0 height or 0 width.
**Assertions:** There are none but it is also harder to think of them for an image class like this.

Bcrims:
### **EventManager Review**
**Design
Informative Identifiers:** The names used are clear and meaningful.
**Avoid Magic Numbers:** There are no arbitrary numbers, time values and counts are handled dynamically.
**Intuitive and Functional Interface:** The interface is straightforward with functions to start, stop, pause, and resume events. Although the use of threads and atomic variables adds some complexity, it is necessary for managing clock-based events
**Informative Comments:** The comments are generally helpful and explain the purpose of major functions
**Consistent Style:** The coding style is consistent throughout, with uniform naming, spacing, and indentation.
**Visual Appearance:** The code is neatly formatted and easy to read, with good use of whitespace and clear separation between functions.

Implementation
**No Warnings During Compilation:** The code compiles without warnings
**Proper Uses of Namespaces:** All code is encapsulated within the cse namespace which helps avoid naming conflicts.
**Adheres to the C++ Standard:** The code follows modern C++ practices and avoids undefined behavior.
**Appropriate Modularity:** The functionality is well segmented into helper functions such as AdvanceTime and TriggerEvents, making the code modular and easier to maintain.
**Uses Modern C++ Constructs:** The use of std::atomic, std::thread, and std::optional shows good adherence to modern C++ practices.
**Uses Standard Library Algorithms:** While the code primarily utilizes container methods, the standard library is used effectively for its needs.
**Pointer Handling:** No raw pointers are used, which minimizes memory management risks.
**Const Correctness:** Accessor functions are marked as const where needed, though additional parameters could be marked const to further clarify intent.

**Tests and Error Checking
Unit Testing:** The tests cover the primary functionalities of the EventManager, though more tests for invalid inputs and edge cases could be beneficial.
**Edge Cases:** Some edge cases are addressed in the tests, but further tests—for example, handling an empty event queue—would enhance reliability.
**Assertions:** Assertions are used in key functions to enforce that only valid event IDs are processed, which helps catch errors early.
**Error Handling:** The class handles error conditions by returning false or empty optionals when events cannot be processed

Overall, the EventManager class is well-designed and effectively implements an event-handling system using modern C++ features.

KollinBartz (commented):
Reviewing the TextBox class
Design
Informative identifiers - Your strings are well worded
Avoids “magic numbers” - You have multiple magic numbers inside the formattedtext.hpp
Intuitive and functional interface - Well done here
Informative comments - Very well written comments
Consistent style - The style is inconsistent with “#pragma once” used in some files and not in others. You also list all of your private variables 2nd and the rest of your group list them first and then public.
Visual appearance - Your code is easy to follow.
Implementation
No warnings during compilation - 
Proper uses of namespaces - You use proper namespaces.
Adheres to the C++ standard - The code compiles and I do not see immediate undefined behavior.
Appropriate Modularity - Modularity looks good.
Uses modern C++ constructs - As of what we learned in class I believe you do.
Uses standard library algorithms - I don’t see any need for them.
Pointer handling - Yes.
Const correctness - They will be once floating points are implemented.
Testing and Error Checking
Unit testing - The unit testing is very good
Edge cases - I don’t see any asserts or edge cases but uncertain as to what they would be 
Assertions - I don’t see any asserts or edge cases but uncertain as to what they would be
Error handling - I would like to see some sort of error handling or asserts. Maybe even some niche cases like overflow or memory management

Group-09/TextBox/FormattedText.hpp
```diff
@@ -0,0 +1,173 @@
...
+  /**
+   * @brief Set the Type object.
+   *
+   * @param newType
+   */
+  void setType(TextType newType) { type = newType; }
+
+  // Setters for font, fontSize, and color.
+  void setFont(const std::string& newFont) {
+    assert(!newFont.empty() && "Font name should not be empty");
```
mercere99, line 113: These are good uses of assert.

Group-09/TextBox/FormattedText.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include <algorithm>
+#include <sstream>
+
+namespace cse {
+/**
+ * @brief Function to escape special characters in raw text.
+ *
+ * @return std::string
+ */
+std::string FormattedText::getEscapedText() const {
```
mercere99, line 19: Very useful conversion!  And it will make your life easier, rather than having to worry about the HTML being processed correct each time you use it.

Group-09/TextBox/TextBox.hpp
```diff
@@ -0,0 +1,192 @@
...
+   */
+  TextBox() = default;
+
+  /**
+   * @brief Construct a new TextBox.
+   *
+   * @param boxID If empty, an ID is auto-generated.
+   * @param config The configuration for the TextBox.
+   */
+  TextBox(std::string boxID, const TextBoxConfig& config)
```
mercere99, line 47: This is a nice technique to help make sure that all input values have associated names.  Much better than overloading your function with too many parameters.

Group-09/TextBox/TextBox.hpp
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  /**
+   * @brief Copy constructor.
+   */
+  TextBox(const TextBox& other)
+      : id(other.id),
+        content(other.content),
+        width(other.width),
+        height(other.height) {}
```
mercere99, line 71: Since you are just copying all of the value over, this code can be simplified to:
```cpp
TextBox(const TextBox &) = default;
```
Similar for the move constructor and the copy and move operators below.

Group-09/WebLayout/WebLayout.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+             textBoxDiv.innerHTML +=
+                 '<p style="position: absolute; left: ' + x + 'vw; top: ' + y
+                     + 'vh; border: 1px solid black; margin: 0; border-radius: 5px; height: ' + height + 'vh; width: '
+                     + width
+                     + 'vw;">'
+                     + msg + '</p>';
+           }
+         }, msg.c_str(), width, height, x, y
+  );
```
mercere99, line 111: Wow -- this looks like a really nice, clean way of handling this function.  Same for images below.

mercere99 (approved):
Very nice start to a solid collection of web classes and event management.

tests/Group-09/TestEventQueue.cpp
```diff
@@ -0,0 +1,165 @@
...
+  EventQueue eq;
+  Event e2(1, 1, "data");
+  Event e3(3, 3, "data");
+  Event e4(4, 4, "data");
+  Event e1_updated(0, 100, "data");
+  eq.add(e2);
+  eq.add(e3);
+  eq.add(e4);
+  REQUIRE_THROWS(eq.update(e1_updated));
+}
```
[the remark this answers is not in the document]
owenhaiar: Priority can be modified, which is what the test case above is checking (order is updated after updating an events priority). IDs cannot be changed

PR #84, opened 2025-02-11
Stroustrup distribution

KollinBartz:
Updated the testing for the distributions class.


tests/Group1/Distibution/Distrubution.cpp
```diff
@@ -0,0 +1,38 @@
...
+  
+  // Test making a uniform distribution
+  distribution.BuildUniform(5);
+  CHECK(distribution.getProb(3) == 0.2);
+  CHECK(distribution.getCumulativeProb(3) == 0.6);
+
+  // Test making a binomial distribution
+  distribution.BuildBinomial(5, .1);
+  CHECK(std::fabs(distribution.getProb(3)-0.0081)<.0001);
+  CHECK(std::fabs(distribution.getCumulativeProb(3)-0.99954)<.00001);
```
LeoSSpecht, line 22: I think using the WithinAbs here is better in order to follow Catch patterns

LeoSSpecht (changes requested):

tests/Group1/Distibution/Distrubution.cpp
LeoSSpecht: Also would add the `0.0001` into a variable so it is not a "random" loose number in the code

PR #89, opened 2025-02-12
Test Cases for WeightedSet class

Jchen3412:


Group-01/WeightedSet/WeightedSet.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once 
+
+#include <vector>
+
+namespace cse498 {
+    class WeightedSet {
+    private:
+        std::vector<size_t> ids;
```
LeoSSpecht, line 8: I think using vectors here will be very inefficient for large WeightedSets.

Group-01/WeightedSet/WeightedSet.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+double cse498::WeightedSet::Weight(size_t id) {
+    auto it = std::find(ids.begin(), ids.end(), id);
+    if (it == ids.end()) {
+        throw std::invalid_argument("ID does not exist");
+    }
+    return weights[std::distance(ids.begin(), it)];
+}
+
+size_t cse498::WeightedSet::At(double position) {
```
LeoSSpecht, line 39: There is a big problem in your implementation. If the order the items are added change, lets say weight 3 is added before weight 1.5, then when using At(1) will trigger the ID of 3.

LeoSSpecht (changes requested):

LeoSSpecht (approved):

KollinBartz (approved):
Looks good!

PR #93, opened 2025-02-12
Hamilton CSV Tests Passed

asifmasood7:
- Added CSVFileTests.cpp under tests in Group 5
- Used bash and "make" the file through terminal
- All Tests Passed


tests/Group-05/CSVFileTests.cpp
```diff
@@ -0,0 +1,149 @@
...
+        std::ofstream out_file(test_file_name);
+        REQUIRE(out_file.is_open());
+        // Write CSV content with a header and two data rows
+        out_file << "Name,Age,Score\n";
+        out_file << "Alice,30,95.5\n";
+        out_file << "Bob,25,88.0\n";
+        out_file.close();
+    }
+
+    // Load the CSV file into a DataGrid
```
MaxKrawec, line 41: Should you use CHECK_THROWS() for an exception? I'm not sure

tests/Group-05/CSVFileTests.cpp
```diff
@@ -0,0 +1,149 @@
...
+    }
+    // Fill second data row
+    {
+        std::vector<cse::Datum> &row2 = data_grid.GetRow(2);
+        row2.push_back(cse::Datum("Diana"));
+        row2.push_back(cse::Datum(32));
+        row2.push_back(cse::Datum(85.5));
+    }
+
+    // Export the DataGrid to a temporary CSV file
```
MaxKrawec, line 114: Same thing as above (Should you use CHECK_THROWS() for an exception? I'm not sure)

.vscode/settings.json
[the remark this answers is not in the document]
asifmasood7: This somehow gets pushed everytime I make any new changes. I will try to not stage this change next time when I push something.

Group-05/src/CSVfile.cpp
```diff
@@ -77,39 +79,55 @@
...
-         }
-       } else {
-         row.push_back(Datum(token));
+    std::ifstream file(file_name);
+    // First, check if the file stream is open.
+    if (!file.is_open()) {
+       throw std::runtime_error("Cannot open file: " + file_name);
+    }
+    // Assert that the file stream is open (for debug builds).
+    assert(file.is_open() && "File must be open for reading.");
```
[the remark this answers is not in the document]
asifmasood7: Yes, because if the file isn’t open the exception is thrown, so the assert is only reached when the file is open - it just documents that assumption in debug builds.

tests/Group-05/CSVFileTests.cpp
asifmasood7: So I think we only need to use CHECK_THROWS (or CHECK_THROWS_AS) in tests where we expect an exception. In this section, we're setting up a valid CSV file, so using REQUIRE to ensure the file is open is appropriate. I think.

tests/Group-05/CSVFileTests.cpp
asifmasood7: Same explanation as above.

tests/Group-05/CSVFileTests.cpp
MaxKrawec: Line 45: Isn't this expecting an exception?

```
catch (const std::exception &e) {
        FAIL("Exception thrown during LoadCsv: " << e.what());
    }
```

Group-05/src/CSVfile.cpp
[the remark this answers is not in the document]
MaxKrawec: Yes, I understand your reasoning. Today he talked about when to use an exception and when to use a assert. I was under the impression that we would use one or the other, but I could have misunderstood him

tests/Group-05/CSVFileTests.cpp
asifmasood7: So in this test we're verifying that a valid CSV file loads correctly, so we don't expect an exception. The catch block is only there to flag an unexpected error by failing the test with a clear message. CHECK_THROWS() is used when you do expect an exception to be thrown. Therefore, I think this approach of using REQUIRE and FAIL in the catch block is correct. I will merge the branch now for tonights due-date sake. And yes I agree with u too so later, if we need to revisit, I will do more research on that part.

PR #95, opened 2025-02-12
EventQueue Asserts and Unit Tests

owenhaiar:
Added asserts to EventQueue peek and pop. Added ID set attribute to EventQueue. Updated unit tests


merryymary (approved):
Looks sublime

graceKC04 (approved):

PR #98, opened 2025-02-12
Added new changes to ComboManager and added the Test cpp file for it.

AmeryJohnson:
Added namespace to ComboManager and changed variable names to match.
Added Test file for ComboManager where all tests are passing at the moment.


Devon-FoxElster (approved):
Looks good to me

PR #99, opened 2025-02-12
Liskov WebLayout | Testing + Image/Textbox integration

merryymary:
@allienello @miabodenhorn I had to add == operator functions to Image and Textbox when I switched from the temporary structs to your classes. Feel free to change it in any way you think is best. I just had to throw it in quickly to get my test cases to work :D

### DIFF
--- a/Group-09/Image/image.hpp
+++ b/Group-09/Image/image.hpp
@@ -25,7 +25,16 @@ private:
 public:
     Image(const std::string& url, int width, int height, const std::string& altText = "");
 
-    /** @return The url of the image */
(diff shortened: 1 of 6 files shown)


merryymary:
Also for context, emscripten doesn't work well with catch so I had to do actual executable functions instead of the makefile testcase automations.

allienello (approved):
Everything looks good! I believe we have to use Catch2 for testing, so may need to integrate that

miabodenhorn (approved):
Really good stuff here

PR #100, opened 2025-02-12
[hamilton] hamilton-datum-tests PR - Datum tests and asserts

MaxKrawec:
Completed:
1) Added tests for the Datum.cpp class (Also created a Makefile for it) 
2) Added asserts throughout the Datum.cpp class
3) Added default constructors for copy, destructor, move, and move assignment. 
4) There was a bug where strings that started with numerical values would convert into a double. For example, "123Hello" would convert into a double with the value of 123. I fixed the issue and these strings should be NaN now.

### DIFF
--- a/Group-05/src/Datum.cpp
+++ b/Group-05/src/Datum.cpp
@@ -4,6 +4,7 @@
  * @author Max Krawec
  */
 
+#include <cassert>
 #include <cmath>
 #include <iostream>
 #include <limits>
@@ -13,7 +14,7 @@
 
 namespace cse {
 
-// TODO - May need to change value, I just picked a random one
+// TODO - May need to change value, I just picked a semi-random one
 static const double kEpsilon = 0.0001;
 
 /**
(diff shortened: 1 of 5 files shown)


tests/Group-05/DatumTests.cpp
```diff
@@ -0,0 +1,247 @@
...
+}
+
+TEST_CASE("AsDouble() functionality", "[datum]")
+{
+  SECTION("When the Datum is a double") {
+
+    SECTION("When Datum contains a string that CAN be converted") {
+      cse::Datum string_test(123.123);
+      string_test.AsDouble();
+      REQUIRE_THAT(string_test.GetDouble().value(), Catch::WithinAbs(123.123, g_epsilon));
```
asifmasood7, line 81: Overall your tests look great and looks like you covered all cases. But just to clarify, in this section we are creating a Datum using a double literal (123.123) rather than a string (e.g. "123.123"). Was that intentional? Using a string here might more clearly test the conversion behavior for string input. This is just an optional comment too where I am just trying to clarify. Rest, your tests look great and I will approve you PR soon.

Group-05/src/Datum.cpp
```diff
@@ -164,6 +182,7 @@ bool Datum::operator==(const Datum &datum) const {
  */
 bool Datum::AreDatumsDouble(const Datum &datum) const {
   if (IsDouble() && datum.IsDouble()) {
+    assert(GetDouble().has_value() && datum.GetDouble().has_value());
     if (GetDouble().has_value() && datum.GetDouble().has_value()) {
       return true;
```
asifmasood7, line 187: This is something **completely optional** - but maybe you can consider simplifying these 2 helper functions "AreDatumsDouble and AreDatumsStrings." Instead of using an if with an assert followed by another check, you could combine all conditions into a single return statement. This might help remove redundant checks and improves clarity.  Again the way you did that is great - so this is something completely on the optional side.

PR #102, opened 2025-02-12
Stroustrup distribution

KollinBartz:
Updated the test cases for Distribution as well as adding it to the makefile. Updated my code inside the distribution file to better represent logically conclusions.


typritchett6 (approved):
Only thing is to add Distribution to the Test_Folders in Makefile.

PR #115, opened 2025-02-15
[hamilton] expressionParser-update PR

MaxKrawec:
Completed:
1) Completely changed how the ExpressionParser class works. Now actually uses a container. 
1.1) Made it a template class so multiple types of container can use it
1.2) Simplified the code since our specs changed.
2) Updated and improved tests.
3) Renamed class

PLEASE make sure this is what the class is suppose to do. It's very different than what you had. I know you talked to the professor about it, so just make sure this is what he wants.


MaxKrawec:
Note: I accidently branched off another branch I was working on, so there are a lot of files that changed. I want you to mainly focus and review these classes: ExpressionParser.cpp, ExpressionParser.h, and ExpressionParserTests.cpp

PR #121, opened 2025-02-17
Added a few more tests to MemoFunction

ephraimbennett:


Devon-FoxElster (approved):

gr005100:
/// Group 10 MemoFunction Review

Design

Informative Identifiers: 
- recursiveFib is used in both memofunction and testmemofunction however you redefine it when it could be reused
- myFunc isnt very clear consider changing the name

Avoids Magic Numbers:
- everything looks good

Intuitive and Functional Interface:
- you could add a return type for clearCache stating if it cleared or failed or clear

Informative Comments:
- everything looks good 

Consistent Style:
- the variable name in the testmemofunction contain a typo fibbonacci instead of fibonacci 

Visual Appearance:
- testPerformance is a tad messy


Implementation

No Warnings During Compilation:
- no errors on my end

Proper Use of Namespaces:
- namespace cse isnt being used through out the whole test file

Adheres to the C++ Standard:
- all good on my end

Appropriate Modularity:
- could use helper function for testperformance 

Uses Modern C++ Constructs:
- consider using std::optional
- consider using constexpr for the Fibonacci test number

Uses Standard Library Algorithms:
- consider std::optional and generate_n

Pointer Handling:
- all good

Const Correctness:
- do const std::function<Output(Input)> func_;
- use const in testperformance


Testing and Error Checking

Unit Testing:
- consider negative input for Fibonacci 

Edge Cases:
- consider large value Fibonacci values

Assertions:
- no asserts 

Error Handling:
- add a case for invalid types for operator
- add a case for memoization to see if its success or not

PR #124, opened 2025-02-17
MemoryFactory Initial Class Final Changes

Devon-FoxElster:
Added functionality to the MemoryFactory Constructor for users to give it an initial value to assign to all objects. I also replaced an exception with an assert and removed the old tests associated with that exception. Finally, I improved upon the comments.


Muelleralex1021 (approved):

ephraimbennett (commented):
sounds good

PR #128, opened 2025-02-18
EventManager: Updated event ID handling and Pause/Resume event handling

graceKC04:
- Modified AddEvent to create and add an event to the EventQueue, and now it returns the event object.
- Updated PauseEvent and ResumeEvent to accept an Event object as their parameter.
- Added new test cases to verify the functionality of pausing and resuming events.
- Updated existing EventManager test cases to accommodate the changes in function parameters.


owenhaiar (approved):

merryymary (approved):

gr005100:
/// Group 9 EventManager Review 

Design

Informative identifiers: 
- replace eM and e with a more informative name

Avoids Magic Numbers:
- you are using .size in the line (event_queue_.size() > 0) you could do !event_queue_.empty() to be more consistent however overall 

Intuitive and Functional Interface:
- you use a informitive assert in PauseEvent and ResumeEvent. However id modify your AdvanceTime since it runs continually as well you should add more to your RepeatEvent since it doesn't handle cases where an event is already marked as repeating 

 Informative Comments:
 - Remove the following comment //Queue is already running

Consistent Style:
- You use camalcase and snake_case try and stick to one

Visual Appearance:
- TriggerEvents is a bit complex you could try and break it down


Implementation

No Warnings During Compilation:
- No errors on my end

Proper Use of Namespaces:
- meets all requirnments

Adheres to the C++ Standard:
- if stopqueue gets called it could cause an issue with advancetime
- your eventmanager dsetructor could have more checks such as clock_thread_

Appropriate Modularity:
- everything is good you could modify pause and resume event logic statements since they are very similar

Uses Modern C++ Constructs:
- good use however you could handle threads 

Uses Standard Library Algorithms:
- only the use of .empty as to size() > 0

Pointer Handling:
- you could use unique or shared pointer however its not not necessarily needed 

Const Correctness:
- use const at the following:
bool PauseEvent(Event &event) - bool PauseEvent(const Event &event)
bool ResumeEvent(Event &event) - bool ResumeEvent(const Event &event)
bool RepeatEvent(cse::Event &event, int time_interval) - bool RepeatEvent(const cse::Event &event, int time_interval)

Testing and Error Checking

Unit Testing:
- consider adding a test for clock time and running for eventmanager
- consider adding tests for the eventmanager destructor such as thread joining
- add test cases for stop and start queue

Edge Cases:
- add a test for a empty queue
- add tests for pause and resume events
- add a test case for addevent to consider negative time (niche)

Assertions:
- everything looks good

Error Handling:
- consider adding error handeling for pause and resume event as to just asserts
- if addevent has a negative time it will cause undefined behavior
- startqueue doesnt check if the clockthread has started and for stopqueue to check if the thread has started before joining

PR #131, opened 2025-02-18
Stroustrup dev branch

LeoSSpecht:
Includes classes:
- Graph
- GraphPosition
- Distribution
- Random
- WeightedSet


tests/Group1/Random/Random.cpp
nigamsa1, line 1: Very thorough tests! If you wanted to you could add more types into the template testing like longs and shorts but overall looks good!

Group-01/WeightedSet/WeightedSet.cpp
```diff
@@ -0,0 +1,52 @@
+#include "WeightedSet.hpp"
+#include <algorithm>
+#include <stdexcept>
+
```
DominikLeisinger, line 4: First thing I notice is no comments on any of the functions stating their purpose, arguments, return value

Group-01/WeightedSet/WeightedSet.cpp
```diff
@@ -0,0 +1,52 @@
...
+void cse498::WeightedSet::Add(size_t id, double weight) {
+    if (weight < 0) {
+        throw std::invalid_argument("Weight must be a positive value");
+    }
+    ids.push_back(id);
+    weights.push_back(weight);
+}
+
+void cse498::WeightedSet::Remove(size_t id) {
+    auto it = std::find(ids.begin(), ids.end(), id);
```
DominikLeisinger, line 14: Could make this variable name more explanatory, like 'requested_remove' or just 'item'

Group-01/WeightedSet/WeightedSet.cpp
```diff
@@ -0,0 +1,52 @@
...
+    if (it == ids.end()) {
+        throw std::invalid_argument("ID does not exist");
+    }
+    size_t i = std::distance(ids.begin(), it);
```
DominikLeisinger, line 18: Same here with more explanatory variable name, like 'dist_to_item' or a comment nearby explaining what is happening

Group-01/WeightedSet/WeightedSet.cpp
DominikLeisinger, line 4: Indentation and appearance for your code is great. Clean, no unnecessary code and format is consistent.

Group-01/WeightedSet/WeightedSet.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once 
+
+#include <vector>
+
```
DominikLeisinger, line 4: Clean code and consistent format,  just need comments describing the class and private member variables

tests/Group1/WeightedSet/WeightedSet.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+TEST_CASE("Test WeightedSet::Add", "[base]")
+{
+    cse498::WeightedSet ws;
+    CHECK_NOTHROW(ws.Add(1, 1.0));
+    CHECK_NOTHROW(ws.Add(2, 2.0));
+    CHECK_NOTHROW(ws.Add(3, 3.0));
+}
+
+TEST_CASE("Test WeightedSet::Remove", "[base]")
```
DominikLeisinger, line 14: If an item is added with the same id as another item, how will you differentiate between the two when removing?

Group-01/WeightedSet/WeightedSet.cpp
```diff
@@ -0,0 +1,52 @@
+#include "WeightedSet.hpp"
+#include <algorithm>
+#include <stdexcept>
+
+void cse498::WeightedSet::Add(size_t id, double weight) {
```
DominikLeisinger, line 5: if the function is passed a negative number for 'id', underflow will occur and give you a very large number for this variable. Unsure how it will affect your code, but will most likely result in an unexpected number for 'id'

tests/Group1/WeightedSet/WeightedSet.cpp
```diff
@@ -0,0 +1,62 @@
+#include "../../../third-party/Catch/single_include/catch2/catch.hpp"
+#include "../../../Group-01/WeightedSet/WeightedSet.hpp"
+
```
DominikLeisinger, line 3: For functions like Add, Remove, and Weight, you will need to handle if two items have the same id.
Add() - Either not allow two items of the same id or figure out some way for your class to differentiate?
Remove()- If an item is added with the same id as another item, how will you differentiate between the two when removing?
Weight() - Which weight will this function return if two items have same id?

iamrb1:
### Code Review for Distribution - Rahul Baragur

**Design**
_Informative identifiers._ 
Throughout the Distribution class, the names of member variables and functions are relatively clear to understand for a reader. I would suggest renaming any member variable that contains “probs” to “probability” as it is clear to readers who are not familiar with binomial or uniform distributions in general. Additionally, I believe that the name would not be too verbose. That way, the average person can understand your code. With regards to your methods, it is easy to understand building the distributions and what methods are “getters”.

_Avoids “magic numbers”._
Unfortunately, there are some “magic numbers” scattered throughout your code. No worries though, it is relatively easy to change the functionality of your code to remove them altogether. For example, there are if statements that contain numbers such as if (prob > 1.0 || prob < 0.0 We can set constexpr double (your var name here) = 1.0 and the same for 0.0. In that way the loop can remove 1.0 and 0.0 and replace it with your specified constants. There are many lines that have 1.0 and 0.0 or 1 and 0 scattered throughout so please add constants as soon as possible.

_Intuitive and functional interface._ 
The class exposes necessary functionality and adheres to OOP rules. Member variables are private and methods are public with getters existing to access member variables. The class is straightforward, as it just calculates distributions and gives the data to the user. I do think that having default arguments for parameters is something you could add. For example in BuildBinomial method, you could have trials default to 1 if not specified within by the user. See if that is something you can add.

_Informative comments._ 
I feel like there are not a lot of descriptive comments for this class. For instance, I would like to see an explanation of why a method works and how the algorithm operates from a general standpoint. Consider adding docstrings in your code before the methods that tell the reader what each parameter is and what the method does. For a reader that is new to your code, it is worth helping them out by explaining certain looping functionality and what is happening with a comment that details your implementation choices on line 55 for BuildUniform in Distribution.hpp. 

_Consistent style._ 
After comparing your code with other code styles within your group, it seems to me that there are certain differences that you may want to change. For example, it seems to me that the rest of your group seems to use snake_case for member variables (or all lowercase) when you utilize camelCase for yours. Consider changing cumulativeProbs to another name that follows your group’s standard (I.E cumulative_probability). Additionally, consider changing getProb to GetProbability (I made a suggestion to rename probs to probability earlier in this document) and getCumuativeProb to GetCumulativeProbability. All methods are in line with the style chosen by your group. Good work!

_Visual appearance._ 
In terms of whitespace, it seems that everyone in your group differs within the implementation of your class after writing namespace cse { one line after or just right after). Other than that, it seems that your class adheres to clang format and is well written and structured. 

**Implementation**
_No warnings during compilation.._
Code compiles with no warnings with the -Wall and -Wextra flags. Perfectly implemented!

_Proper uses of namespaces._ 
The code is within namespace cse and there are no uses of using namespace std within the hpp file!

_Adheres to the C++ standard._ 
The code compiles and runs with no undefined behavior when running the cpp file. The code adheres to the C++ standard.

_Appropriate Modularity._ 
I think that a helper function could be added to calculate binomial probability so that the code looks cleaner within BuildBinomial. Other than that, everything adheres to this rubric standard. The getters share some similar code; it could be possible to make it into one function to reuse code and add a parameter to specify what member variable to get.

_Uses modern C++ constructs._ 
Constexpr can be used for the creation of constants as mentioned earlier in this document. I also believe that for loops can be removed altogether (see below). Other than these two things the code does not have many use cases for modern C++ constructs. 

_Uses standard library algorithms._ 
The code utilizes std::tgamma which is great to see for the creation of the algorithm of binomial distribution. In terms of feedback, certain std algorithms such as std::accumulate or std::partial_sum can be utilized to replace your for loops (I would point to uniform distribution for this one -> you can use an std library algorithm here). Try avoiding loops if you can unless it is not possible.

_Pointer handling._ 
No pointers in this code whatsoever.

_const correctness._ 
The member variables correctly use const so great job! You should consider changing your getters and mark them const since they do not change the state of the class.

_Unit testing._ 
As of right now, there are tests for the functionality of the class for BuildUniform, BuildBinomial, error checks for both, getter functionality and correct functionality. These tests look great and well organized. I would add some additional functionality; however. Consider adding two additional test cases before the binomial and uniform distribution tests that are separate from functionality and test the Constructors, and the getters by themselves. This would ensure organization and a bottom to top testing approach that would look nice. 

_Edge cases._ 
Testing covers edge cases such as out_of_range, runtime errors for nonsensical values, and empty values. Overall great job making tests for these edge cases.

_Assertions._ 
There are no assert statements utilized within the code whatsoever. I think that an assert can be added for ensuring trials >= 1 or prob being >= 0 and <= 1.0 instead of throwing exceptions. Consider changing the functionality of unnecessary exceptions to utilizing asserts.

_Error handling._ 
You make sure to return correct exceptions such as out of range or runtime error when necessary so good job!

Group-01/Distribution/Distribution.hpp
```diff
@@ -0,0 +1,95 @@
+#pragma once
```
iamrb1, line 1: Code Review for Distribution left as comment on pull request -> Associated Changes detailed within code review

iamrb1 (commented):
Please see associated comment above for code review.

ze-ntsh:
# Code Review for Random

## **Design**

### **Informative Identifiers**
- The identifiers are generally clear and meaningful. For example:
  - `GetDouble`, `GetInt`, and `P` are descriptive of their functionality.
  - `generator` is a good name for the random number generator.
- The template function `Get<T>` is appropriately named, and the use of `min` and `max` is intuitive.
- **Suggestion**: Consider renaming `P` to something more descriptive, like `CheckProbability` or `EvaluateProbability`, to make its purpose clearer.

### **Avoids "Magic Numbers"**
- The code avoids magic numbers effectively. For example:
  - The bounds in `GetDouble`, `GetInt`, and `P` are passed as parameters, avoiding hardcoded values.
  - The `probability` parameter in `P` is checked against `0.0` and `1.0`, which are reasonable bounds for probabilities.
- **Suggestion**: None needed; this is well-handled.

### **Intuitive and Functional Interface**
- The class interface is intuitive and functional:
  - The methods are straightforward and easy to use (`GetDouble`, `GetInt`, `P`, `SetSeed`).
  - The template function `Get<T>` is a nice addition, providing flexibility for numeric types.
  - Member variables are private, which is good encapsulation.
- **Suggestion**: Consider adding a default constructor for `Random` that initializes the generator with a random seed (e.g., using `std::random_device`). This would make the class more flexible.

### **Informative Comments**
- The code lacks comments, which could make it harder for someone new to understand the purpose of certain methods or design choices.
- **Suggestion**:
  - Add comments to explain the purpose of the class and its methods.
  - For example, explain why `P` uses `0.0` and `1.0` as bounds and what the method is intended to do.
  - Add a comment to the `Get<T>` template to clarify its purpose and constraints.

### **Consistent Style**
- The code follows a consistent style:
  - Naming conventions are consistent (`camelCase` for methods and `snake_case` for variables).
  - Indentation and brace placement are consistent.
- **Suggestion**: None needed; the style is consistent and clean.

### **Visual Appearance**
- The code is visually well-structured:
  - Proper use of whitespace and indentation.
  - Lines are not overly long, making the code easy to read.
- **Suggestion**: None needed; the visual appearance is good.

---

## **Implementation**

### **No Warnings During Compilation**
- The code is well-written and compiles without warnings under `-Wall` and `-Wextra`.

### **Proper Uses of Namespaces**
- The code is correctly placed within the `cse` namespace.
- The use of `using` statements is limited to specific standard library components, which is acceptable.
- **Suggestion**: None needed.

### **Adheres to the C++ Standard**
- The code adheres to the C++ standard:
  - No undefined behavior is apparent.
  - The use of `static_assert` and `assert` ensures valid usage.
- **Suggestion**: None needed; the code is standards-compliant.

### **Appropriate Modularity**
- The code is modular:
  - The `Random` class encapsulates all random number generation functionality.
  - The template function `Get<T>` avoids code duplication for different numeric types.
- **Suggestion**: None needed; modularity is well-handled.

### **Uses Modern C++ Constructs**
- The code uses modern C++ constructs effectively:
  - `static_assert` ensures type safety in the template function.
  - `if constexpr` is used to handle different types at compile time.
- **Suggestion**: None needed; modern C++ is used appropriately.

### **Uses Standard Library Algorithms**
- The code uses the standard library effectively (e.g., `mt19937`, `uniform_int_distribution`, `uniform_real_distribution`).
- **Suggestion**: None needed; standard library usage is appropriate.

### **Pointer Handling**
- The code does not use pointers, so this criterion is not applicable.
- **Suggestion**: None needed.

### **const Correctness**
- The code could benefit from better `const` correctness:
  - Member functions like `GetDouble`, `GetInt`, and `P` do not modify the state of the class and should be marked `const`.
  - For example:
    ```cpp
    double GetDouble(double min, double max) const;
    int GetInt(int min, int max) const;
    bool P(double probability) const;
    ```
- **Suggestion**: Mark non-modifying member functions as `const`.

---

## **Testing and Error Checking**


### **Unit Testing**
- **Positive**: Tests cover all public methods (`GetInt`, `GetDouble`, `P`, `SetSeed`, `Get<T>`), and are well-organized into logical sections.
- **Missing**: No tests for `SetSeed` edge cases or constructor behavior with extreme/edge-case seed values.


### **Edge Cases**
- **Positive**: Tests cover boundary values (e.g., `P(0.0)` and `P(1.0)`) and negative ranges for `GetInt`.
- **Missing**: No tests for invalid inputs or edge cases like `min == max`.


### **Assertions**
- **Positive**: The class uses `assert` internally to enforce invariants (e.g., `max >= min`), and tests implicitly verify that these work.

### **Error Handling**
- **Positive**: The class uses exceptions (e.g., `std::out_of_range`) for recoverable errors, and tests implicitly verify valid inputs work.
- **Missing**: No explicit tests to verify exceptions are thrown for invalid inputs (e.g., `min > max` or invalid probability values).

## **Summary of Suggestions**
1. Add comments to explain the purpose of the class and its methods.
2. Rename `P` to something more descriptive, like `CheckProbability`.
3. Add a default constructor for `Random` that initializes the generator with a random seed.
4. Mark non-modifying member functions as `const`.
5. Add some tests for invalid/extreme inputs.

---

Thisisnotgoingtowork:
## Code review for the WeightedSet

Over all the requirements are fulfilled, there are just some small suggestions.  

### Design

1. The class and member's name in overall are pretty clear, and all in the right format and capitalized, which is great. It just that the **At()** function may be a little confusion for a starter to understand quickly what it is going on. 
2. Even though there are  (position < 0) , but I do think this is domain-specific checks, so it should be acceptable, but it would be better, just better to name it?
3. Do you want to allow duplicates of the same **id**? If not, you might want to check and throw or ignore in **Add**.
4. Maybe add some more brief “what/why” comments above each function explaining the expected behavior, like for **At**, to make it more understandable? I know you have a Readme file, but that didn't really explain the behavior of the functions.

### Implementation

1. Under -Wall -Wextra, you might get a “control reaches end of non-void function” warning in At(double position) since there’s no return after the for-loop. To fix that, you may want to add a throw or return 0 or something like that.
2. In the test file, you refer to cse498::WeightedSet. Possibly a mismatch with cse::WeightedSet? Make sure the namespace is consistent throughout.
3. TotalWeight(), Weight(size_t), and At(double) do not modify WeightedSet; they should be marked const. Also you will also need to mark any helper functions used within them as const.

### Testing & Error Checking

1. Good that it cover all the functions and tested them, still, you might, just might want to add some more edge testing? like direct test for attempting to add a negative weight? Or like operations on an *empty* WeightedSet object? 

And that all I can find potential fix for this class, other than those, this class I think had met the requirements. 
Thank you.

Group-01/GraphPosition/GraphPosition.cpp
hawkinsma, line 1: GraphPosition looks great so far, I honestly couldn't find much to critique.

hoggbram (commented):
Group 1 – Graph:
Design
•	Informative identifiers. Are names of variables, functions, types, etc, clear and helpful for a reader to understand the code. For example, if you need to track the number of students in a class, good variable names might be num_students or student_count, whereas you don’t want something too short to be confusing (like N or num_s) or too verbose that it’s annoying (like total_number_of_students_in_class).
Yes, the names of identifiers are clear and helpful
•	Avoids “magic numbers”. Similar to above, you should use constant variables with informative names instead of arbitrary numbers throughout your code, and instead set constants. For example, I have a virtual CPU class with 6 virtual registers, but rather than using “6” throughout, I set static constexpr size_t NUM_REGS = 6; at the top of the class to make it clearer what the 6 refers to (and, in this case so that I can change it if needed.)
Does avoid magic numbers, as none are present in the first place
•	Intuitive and functional interface. Does the class expose only the necessary functionality, and is it easy to use correctly while hard to misuse? Are member variables private? Is there an obvious function to use for common tasks, with relevant default arguments for parameters? Are any of the functions misleading, like potentially confusing uses of operator overloading? (Feel free to provide suggestions on how to improve the interface for better usability)
Overloading is done well/usefully. The function names are good, and parameter names are good for the most part, having comments would be helpful (although adding comments wasn’t a requirement yet, so not really a criticism). 
•	Informative comments. Do comments help someone new to the code understand it more rapidly? Did the coder properly avoid redundant or obvious comments that clutter the code (e.g., ++x; // increments x). In general, comments should focus on why code is helpful, rather than belabor what that code does (though with complex code, the latter is useful too.)
No comments added so far. Majority of the code is self explanatory, although comments are never harmful to have.
•	Consistent style. Are style choices (like naming schemes, brace placement, etc) used in a consistent and informative manner throughout the code? Is there consistency in how different naming conventions are used? (e.g., camelCase, snake_case, ALL_CAPS, or PascalCase, or prefixes and suffixes, like m_ at the beginning of a member variable.).
Yes the style is consistent.
•	Visual appearance. Is the code easy to scan through and visually well structured with good use of whitespace, indentation, and line breaks? Do there appear to be consistent limits on line widths?
Yes it is clear to see and no obvious flaws in readability. Only change I might suggest is a space between function declarations in the header for easier readability.
Implementation
•	No warnings during compilation. Does the code compile with the -Wall and -Wextra flags without producing warnings? One option is to add -Werror to your list of options, which turns all warnings into errors and thus halts compilation when one occurs.
There are no errors when ran with -wall and -werror and wextra flags.

•	Proper uses of namespaces. Is all code within namespace cse { ... } ? Does the code avoid the generic using namespace std as well as avoiding ALL namespace pollution in header files that will spread to users of those files.
Yes it is in the namespace

•	Adheres to the C++ standard. Does the code avoid undefined behavior? If possible, make sure the code compiles and runs properly on different compilers and different platforms.
I can’t find any undefined behavior here

•	Appropriate Modularity. Are common procedures grouped into useful helper functions? Are helper classes used where needed? Is unnecessary code duplication avoided?
Yes helper classes and functions (structs, getters, setters) are all used well.

•	Uses modern C++ constructs. Where applicable, does the code use range-based for loops, std::optional, or other modern C++ techniques? (In the future we will also look for proper use of templates, move semantics, lambdas, and constexpr, but only once we cover those topics).
Yes it does use some, such as std::optional.

•	Uses standard library algorithms. Does the code avoid extraneous loops when equivalent functionality exists in the standard library? For example, std::fill will fill a vector, std::remove_if will remove elements from a container based on a condition, or std::is_sorted will tell you if the values in a container are in order. See https://en.cppreference.com/w/cpp/algorithm for a full list (it’s LONG, but well worth being familiar with).
The only loop used is find, which is good to use as this is a perfect use case.

•	Pointer handling. Is memory management done carefully, with either smart pointers (std::unique_ptr, std::shared_ptr) or, only if needed, careful use of raw pointers with new and delete?
There aren’t any pointers used other than iterators, which should be destroyed after being used.

•	const correctness. Are member functions to const if they don’t change the state of the class? Are parameters (especially reference parameters) marked const if they don’t change inside the function?
All of the functions do change the class. I believe a decent amount of the string parameters could be marked as const, if I’m understanding the functionality correctly.

Testing and Error Checking
•	Unit testing. Are unit tests present for all public member functions (including constructors) and well organized? Do they trigger cases that will exercise most private portions of the class’ code?
Yes looks good to me. There is only one warning when running the tests: it says edge.cpp is catching a runtime error as a value.

•	Edge cases. Does testing cover most edge cases, as possible? For example, for pointers, does it test null pointers (if they are allowed)? Does it test for empty containers being used? Does it test for 0 or other boundary conditions on parameters?
Yes it tests multiple different cases for some functions

•	Assertions. Are assert statements used properly throughout the code to enforce invariants and ensure a valid state? (for example, checking that valid arguments are passed into functions.) Note that asserts should be used only to identify conditions that should never occur in properly working code; if code is expected to recover, error handling must be performed.
Yes there are thrown errors scattered throughout the implementation

•	Error handling. Are error conditions that can occur during correct execution (such as resource limitations, files not available, or division by zero) checked for and reported back with either exceptions, return conditions, or some other technique?
Yes it is, they seem like they are used in the test files as if they were actually implemented

vernonriley:
This is for Graph class and everything it encaptures.

Design: 
This code is great, everything is formatted extremely well and it is all very easy to read/understand. I also like how you have to do lists at the top of each files when you need to add more things. The only thing I would consider adding is more comments, you do have a lot of code and it could be easy to get lost in without some good comments. I would also add some comments for all of your test cases, they are a bit vague.

Implementation:
I got no warnings when I compiled, you properly use the cse namespace, and you have used a great deal of things in standard library. All of your consts seem to be correct as well (that took a while). There is not much to say on this subject, you did a great job with the implementation. 

Testing: The testing is very thorough, it looks like you have everything tested from your functions in some way or another. It also seems like the edge cases are tested. You also might want to consider adding some more assert statements, I noticed you had a few in Edge.hpp, but that was all.

garciaow:
Code Review for Distribution
Owen Garcia

# Design
**1. Informative identifiers:**
Overall, great job. All names were informative, and I was able to understand what each function did and what each variable was for.

**2. Avoids "magic numbers":**
The code does use 1.0 a few times in the BuildBinomial function and 0.0 in one of its if statements; those could be replaced with constants. There's already a private member variable for the latter that you can use.

**3. Intuitive and functional interface:**
I found this to be pretty clean. 

**4. Informative comments:**
To be completely honest, the code is clean enough and I was able to understand pretty much everything with the comments that were already present. This may come off as redundant, but I think I would suggest just adding as much as you can (maybe a note about why some values are being set equal to something) even if they seem obvious.

**5. Consistent style:**
I did notice that the getter functions used camel case while everything else capitalized each word.

**6. Visual appearance:**
I thought this was pretty clean. 

# Implementation
**1. No warnings during compilation:**
The code compiled for me, and I had no warnings, issues, or instances of undefined behavior. Well done.

**2. Proper uses of namespaces:**
The code is inside namespace cse and there's no sign of namespace std. All good here.

**3. Adheres to the C++ standard:**
The code follows the C++ standard.

**4. Appropriate Modularity:**
The getter functions appear to repeat a lot of their code; I think it would be pretty simple to merge them together and to just add a parameter to specify which probability you want to return.

**5. Uses modern C++ constructs:**
I like the use of std::vector::assign, though I think you might benefit from using constexpr for your variables.

**6. Uses standard library algorithms:**
The code uses these for the binomial distribution, though isn't there an std::binomial_distribution out there? I'd suggest looking into it if you haven't already; if there's some reason as to why it wouldn't work that I overlooked, please disregard this.

**7. Pointer handling:**
The code didn't use any pointers.

**8. Const correctness:**
I think the getters could be made const since they don't modify the objects.

# Testing and Error Checking
**1. Unit testing:**
I think the tests are done pretty well. I don't know what more to suggest.

**2. Edge cases:**
I think these edge cases are pretty thorough; invalid input values, out-of-range indices, and attempts to generate probability without a distribution are all accounted for.

**3. Assertions:**
There were no assertions in the code. I think you could do something to ensure that the probability remains between 0.0 and 1.0.

**4. Error handling:**
See the edge cases section; should something go wrong, everything I can think of seems to be accounted for. Well done.

Group-01/Random/Random.hpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+int Random::GetInt(int min, int max) {
+  assert(max >= min);
+  uniform_int_distribution<int> distribution(min, max);
+  return distribution(generator);
+}
+
+bool Random::P(double probability) {
+  assert(probability >= 0.0);
```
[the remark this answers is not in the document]
mercere99: This is a reasonable idea to consider, but values like zero (and, in most cases, one) can usually be put in the code directly as long as it is clear what they mean.

Group-01/Distribution/Distribution.hpp
```diff
@@ -0,0 +1,102 @@
...
+    void BuildPowerLaw(const int);
+    double GetProb(const int);
+    double GetCumulativeProb(const int);
+    double ProbCalculator(const std::vector<double>, const int trials);
+  };
+
+  // Fills the probs and cumulativeProbs vectors with binomial probabilities
+  void Distribution::BuildBinomial(const int trials, const double prob) {
+    if (trials < minTrials) {
+      throw std::runtime_error("Must have at least 1 datapoint"); // Use the cse::assert in future
```
mercere99, line 34: I understand that one reviewer suggested that you remove magic numbers, but I think the code would be clearer with those values here when they are minimal limits like this that don't have a special meaning.  I'll make sure to clarify in class when you would call something a "magic number", but zero and one rarely count.

Also, these exceptions are more of a case where asserts would fit better.  You should never call BuildBinomial with numbers out of ranges; it's not merely something where you might want to catch and exception and fix things.

mercere99 (approved):
Many comments throughout the code, but overall all 5 classes are in a good place, and can be merged without messing up anyone else's code.

The only changes outside of the group's own directories was to .gitignore, but a reasonable change for full project.

Group-01/GraphPosition/GraphPosition.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+  enum class TraversalMode { DFS, BFS, RANDOM, WEIGHTED };
+
+  class GraphPosition {
+  private:
+    [[maybe_unused]] const Graph &graph;
+    std::shared_ptr<Vertex> currentVertex;
+    std::set<std::shared_ptr<Vertex>> visitedVertices;
+    std::vector<std::shared_ptr<Vertex>> traversalPath;
+    TraversalMode traversalMode = TraversalMode::DFS; // Default to DFS
```
mercere99, line 21: You should put a `[[maybe_unused]]` before this definition since it's not currently used (but clearly intended for future development).

tests/Group1/Graph/Graph.cpp
```diff
@@ -0,0 +1,196 @@
...
+  CHECK(graph.IsConnected(v4, v5));
+  CHECK(!graph.IsConnected(v5, v4));
+  CHECK(e2.use_count() == 1);
+
+  // Testing removing Edges
+  auto v4_v5_edge = graph.GetEdge(v4->GetId(), v5->GetId());
+  CHECK(!v4_v5_edge.expired());
+  {
+    auto e = v4_v5_edge.lock();
+    REQUIRE_THAT(e->GetWeigth(), WithinAbs(2, cse_test_utils::FLOAT_DELTA));
```
mercere99, line 55: `GetWeight()` is misspelled and broken...

tests/Group1/Graph/Graph.cpp
```diff
@@ -0,0 +1,196 @@
...
+  CHECK_THROWS_AS(graph.RemoveEdge(e2), std::out_of_range);
+  CHECK_THROWS_AS(v4->GetEdge(v5), std::runtime_error);
+
+  // No bidirectional edges
+  auto e3 = graph.AddEdge("id1", "id2", 2);
+  CHECK(graph.IsConnected(v1, v2));
+  CHECK(!graph.IsConnected(v2, v1));
+  {
+    auto e = e3.lock();
+    REQUIRE_THAT(e->GetWeigth(), WithinAbs(2, cse_test_utils::FLOAT_DELTA));
```
mercere99, line 69: `GetWeight()` is misspelled here too.

Group-01/Graph/Edge.hpp
```diff
@@ -0,0 +1,77 @@
...
+          "from", SerializableProperty([this](std::ostream &s) { s << from->GetId(); }, [](const std::string &) {}));
+      properties.emplace_back(
+          "to", SerializableProperty([this](std::ostream &s) { s << to->GetId(); }, [](const std::string &) {}));
+      return properties;
+    }
+    void SetId(std::string newId) override { id = newId; };
+
+  public:
+    Edge() = delete;
+    virtual ~Edge() = default;
```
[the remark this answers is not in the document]
LeoSSpecht: Fixed

Group-01/Graph/Graph.cpp
```diff
@@ -0,0 +1,302 @@
...
+   * @throws out_of_range if vertex doesn't exist
+   */
+  std::shared_ptr<cse::Vertex> cse::Graph::RemoveVertex(std::string const id) {
+    auto it = vertices.find(id);
+    if (it == vertices.end()) {
+      throw std::out_of_range("Vertex does not exist: " + id);
+    }
+    auto removedVertex = it->second;
+    vertices.erase(it);
+    return removedVertex;
```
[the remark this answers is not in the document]
LeoSSpecht: Fixed by not returning anything

Group-01/Graph/Graph.cpp
```diff
@@ -0,0 +1,302 @@
...
+   */
+  /**
+   * Adds a new vertex to the graph
+   * @param id Unique identifier for the vertex
+   * @param X X-coordinate position
+   * @param Y Y-coordinate position
+   * @return Shared pointer to the created vertex
+   * @throws runtime_error if vertex ID already exists
+   */
+  std::shared_ptr<cse::Vertex> cse::Graph::AddVertex(std::string const id, double X, double Y) {
```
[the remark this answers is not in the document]
LeoSSpecht: I changed the functions mentioned to return a reference.
Internally I would prefer to keep them as shared pointer, as they have some interface with Edge, so I think keeping them shared is better.

Please let me know if you don't think this is ideal!

Group-01/Graph/Graph.cpp
```diff
@@ -0,0 +1,302 @@
...
+    return removedVertex;
+  }
+
+  /**
+   * Validates that both vertices exist in the graph
+   * @param v1_id First vertex ID
+   * @param v2_id Second vertex ID
+   * @throws out_of_range if either vertex doesn't exist
+   */
+  void Graph::ValidateVerticesExist(const std::string &v1_id, const std::string &v2_id) const {
```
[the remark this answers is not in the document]
LeoSSpecht: Fixed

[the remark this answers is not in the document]
LeoSSpecht: Definitely agree.
Fixed.

Group-01/Graph/Graph.cpp
```diff
@@ -0,0 +1,302 @@
...
+  /**
+   * Adds a new vertex to the graph
+   * @param id Unique identifier for the vertex
+   * @param X X-coordinate position
+   * @param Y Y-coordinate position
+   * @return Shared pointer to the created vertex
+   * @throws runtime_error if vertex ID already exists
+   */
+  std::shared_ptr<cse::Vertex> cse::Graph::AddVertex(std::string const id, double X, double Y) {
+    if (vertices.find(id) != vertices.end()) {
```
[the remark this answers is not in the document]
LeoSSpecht: I made the suggested changed for the `if` but kept the exceptions.

In my opinion thinking of an API that could be used for external use, specially thinking of the application that would be a web interface I think exceptions can be useful to handle in JS.
However, I have not used Emscripten a lot, so I could be very very wrong here.

Group-01/Graph/Graph.cpp
```diff
@@ -0,0 +1,302 @@
...
+  }
+
+  /**
+   * Adds a new edge to the graph using vertex IDs
+   * @param v1_id Source vertex ID
+   * @param v2_id Destination vertex ID
+   * @param bidirectional Whether the edge is bidirectional
+   * @return Weak pointer to the created edge
+   */
+  std::weak_ptr<cse::Edge> cse::Graph::AddEdge(std::string const v1_id, std::string const v2_id, bool bidirectional) {
```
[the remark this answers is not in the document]
LeoSSpecht: Changed, I also agree with this

Group-01/Graph/Graph.cpp
```diff
@@ -0,0 +1,302 @@
...
+    return edge_ptr;
+  }
+
+  /**
+   * Gets an edge between two vertices
+   * @param from Source vertex
+   * @param to Destination vertex
+   * @return Weak pointer to the edge
+   */
+  std::weak_ptr<cse::Edge> cse::Graph::GetEdge(std::shared_ptr<cse::Vertex> const from,
```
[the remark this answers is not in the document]
LeoSSpecht: Changed together with the other parts

tests/Group1/Graph/Graph.cpp
LeoSSpecht: Fixed, sorry about that

Group-01/WeightedSet/WeightedSet.cpp
```diff
@@ -0,0 +1,52 @@
...
+    if (weight < 0) {
+        throw std::invalid_argument("Weight must be a positive value");
+    }
```
EvilJuiceBox, line 8: You can consider asserting that ids and weights have the same size before you push to make sure the items line up.

Group-01/Graph/Graph.cpp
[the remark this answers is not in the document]
LeoSSpecht: Done on the advanced class

Group-01/Util/Util.hpp
```diff
@@ -0,0 +1,140 @@
...
+        : handler(h), writer(w) {}
+    SerializableProperty(std::function<void(const std::string &)> h, std::function<void(std::ostream &)> w)
+        : handler(h), writer(w) {}
+  };
+
+  /**
+   * There was some use of AI to write the types and maps
+   * that map properties to handlers
+   */
+  class FileSerializable {
```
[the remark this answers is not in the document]
LeoSSpecht: Will do for advanced classes

PR #135, opened 2025-02-18
[hamilton] Code Review PR

MaxKrawec:
Classes:
1) Datum
2) ReferenceVector
3) CSVFile
4) DataGrid
5) ExpressionParser


greencalen17 (approved):

MaxKrawec:
Note: This is for the code review. DO NOT merge

nstanis21 (commented):
Review for ExpressionParser:
You could make the const std::string values in ParseKey, MakeDivideFun, MakeAddFun, MakeSubtractFun, and MakeMultiplyFun into const std::string & values since you do not make any changes to those values and it should improve performance.

You have unnecessary comments that just pollute whitespace, such as lines 16, 20, 23, 30, 37, 111, 117, 123, 124, 148 in ExpressionParser.cpp and line 80 in ExpressionParser.h, these comments are not describing any advanced functionalities or things added to meet specific guidelines, and none of these comments really fit either criteria.

ParseKey could throw an error from being given a string of the form "{....", as the lack of the closing curly brace will cause the loop in lines 24-27 of ExpressionParser.cpp to increment 'index' to an out of bounds position in 'expression', which the function then may attempt to access in line 29, causing an error. It may be worthwhile to add error handling for 'bad' expressions, with corresponding tests.

Is there some way you might be able to make the symbolTable you give the parser (like you have in your tests) into a member variable of the class? This feels like something the class should be holding to respond to expression inputs instead of also taking it as an input when taking expression inputs. This approach should also improve performance since you are currently having to fully copy the map into memory for each function call, which would not occur if the table were a class object.

matrix373 (commented):
### Review for CSVfile

- There are couple of points that I referred to use more descriptive naming in the review but other than those, naming convention looks great to me.
- Comments and documentation looks great and was helpful for me to understand the code.  Overall visual appearance looks great.
- There wasn't any warnings or errors and all tests are passing for CSVfile.
- I recently found a standard algorithm called std::all_of. As a suggestion, you might want to take a look at for future use. 
- Testing looks promising, I recommend testing more edge cases if possible. For instance, having tests for empty fields or special characters might be useful.

Overall, your code looks excellent and is easy to read or understand. Great job!

Group-05/src/CSVfile.cpp
```diff
@@ -0,0 +1,192 @@
...
+namespace cse {
+
+/**
+ * @brief Trims the whitespaces from both ends of a string.
+ *
+ * @param s The string to trim.
+ * @return The trimmed string.
+ */
+static std::string Trim(const std::string &s) {
+  size_t start = s.find_first_not_of(" \t\r\n");
```
sidb70, line 28: I would recommend making these constant for this set of whitespace characters

Group-05/src/CSVfile.cpp
```diff
@@ -0,0 +1,192 @@
...
+  size_t end = s.find_last_not_of(" \t\r\n");
+  // CITE: Used ChatGPT to write this return statement and to find a method for trimming whitespaces effectively.
```
sidb70, line 30: appreciate the use of references, but consider removing the GPT citations to improve readability. these also do not add to a user's understanding of the codebase.

Group-05/src/CSVfile.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+namespace cse {
+
+/**
+ * @brief Trims the whitespaces from both ends of a string.
+ *
+ * @param s The string to trim.
+ * @return The trimmed string.
+ */
+static std::string Trim(const std::string &s) {
```
sidb70, line 27: Naming of Trim and Sanitize is vague. Perhaps something like "TrimWhitespaces" is a more intentful interface.

Group-05/src/CSVfile.cpp
```diff
@@ -0,0 +1,192 @@
...
+ * @param delimiter The delimiter used in the CSV file.
+ * @return True if the export was successful.
+ * @throws std::runtime_error If the file cannot be written.
+ */
+bool CSVFile::ExportCsv(const std::string &file_name, const DataGrid &grid,
+                        char delimiter) {
+  std::ofstream out_file(file_name);
+
+  // Assert that the output file stream is open and is active in debug mode.
+  assert(out_file.is_open() && "Output file must be open for writing.");
```
sidb70, line 149: in ExportCsv, your assertion to check whether the file is open is before your runtime error, but the order is flipped in the LoadCsv function. consider which would be more appropriate for your specification and maintain consistency throughout your functions

Group-05/src/CSVfile.cpp
```diff
@@ -0,0 +1,192 @@
...
+ * @brief Trims the whitespaces from both ends of a string.
+ *
+ * @param s The string to trim.
+ * @return The trimmed string.
+ */
+static std::string Trim(const std::string &s) {
+  size_t start = s.find_first_not_of(" \t\r\n");
+  size_t end = s.find_last_not_of(" \t\r\n");
+  // CITE: Used ChatGPT to write this return statement and to find a method for trimming whitespaces effectively.
+  return (start == std::string::npos || end == std::string::npos)
```
sidb70, line 31: consider using std::find_if for the Trim function implementation.

sidb70 (commented):
The following feedback report is identifying areas which can be improved. The topics which are not addressed in this review (const correctness, consistent style, etc.) were deemed to be acceptable or excellent. 

### Design

- **Informative identifiers**: in general, the identifiers are clear. However, helper functions like Trim and Sanitize can be made more informative. 
- **Avoids “magic numbers”**: only slight repetition in the Trim function; a constant can be used to represent the set of whitespace characters. 
- **Intuitive and functional interface**. The LoadCsv and ExportCsv are very intuitive interfaces. Given that there is no statefulness, the static methods are appropriate. The helper functions such as Trim and Sanitize can be improved by updating the naming convention. 
- **Informative comments**. many comments are in the code, however, some of them are unnecessary/clutter (referencing gpt or that the developer researched on stackoverflow). consider removing comments which do not clarify something about the code or help understanding.

### Implementation
- **Proper uses of namespaces**. all code is in namespace. however, consider moving the Trim, Sanitize, and IsNumeric helper functions outside of the namespace to avoid pollution. 
- **Uses modern C++ constructs**. Current code has manual for loops. Consider using range based for loops and using iterators. 
- **Uses standard library algorithms**. Could potentially use std::find_if for the Trim helper function.

### Testing and Error Checking

- **Unit testing.** Good coverage of basic functionality.  Consider testing some more coverage of common situations that may occur like loading an empty CSV. Does it correctly load it? also consider testing your helper functions since your member function functionality is dependent on them.

 as mentioned by another reviewer, it seems that the Export function will convert a numeric value of 0.0 into an empty string. this was not checked in the tests. 

- **Edge cases.** More edge cases can be added. What if the rows have different numbers of tokens? what if the csv has only a header row but no data. what if the fields within the csv intentionally have whitespace/newline/special characters?
- **Assertions**. asserts are indeed added in the source code, but ensure that they are used consistently. For example, Export function has an assert statement before the runtime error to check if the file is open, but the order is reversed in Load.

- **Error handling.** consider adding more error cases, such as if the user enters an invalid filepath or tries to export to a readonly file.

Group-05/src/Datum.h
```diff
@@ -0,0 +1,120 @@
...
+   * @return true if the type is a double, false otherwise.
+   */
+  [[nodiscard]] bool IsDouble() const { return std::holds_alternative<double>(value_); }
+
+  /**
+   * Returns the Datum as a variant.
+   * @return the variant value.
+   */
+  [[nodiscard]] std::variant<std::string, double> GetVariant() const { return value_; };
+
```
rahnsieg, line 81: Make sure to add a comment for the rest of the functions here

Group-05/src/Datum.h
rahnsieg, line 1: Overall Feedback:
- Make sure to leave a comment for all of your functions
Overall this looks great!

Group-05/src/Datum.cpp
rahnsieg, line 1: Overall Feedback:
- I would add some more comments to the bodies of your functions to explain some of the code
- Make sure you are testing for all the edge cases in your class
Overall it looks great, especially the fact that you cited any sources you used for code

tdkduong (commented):
Review for DataGrid:
- Coding: Overall, the code solved the requirements pretty well. However, if it is possible, I think it will be better if any function parameter change from double/std::string to Datum, since some might want their DataGrid contains both double and string.
- Testing: The tests are good. However, it can be better if the value are randomly generated in both type of double and strings.

aydinorh:
Review for ExpressionParser:

Design:

- Informative identifiers: I would use more descriptive names instead of name1 and name2(maybe key1, key2). Other than that naming of the variables and functions are good.
- Avoids “magic numbers”: You can use constexpr for "-, +, /, *".  Maybe you can define them as private. I don't see any issue with the magic numbers.
- Intuitive and functional interface: Overall a good interface, very clear public and private. I would suggest making num_map parameter in Evaluate() const& to avoid copies.
- Informative comments: Your Doxygen format looks good. I would remove comments like line 16 and 20 in cpp file.
- Consistent style: Your style choices for function names and variable names consistent throughout the code. 
- Visual appearance: Your code visually appears good and easy to understand.

Implementation:

- No warnings during compilation: There was no warning during compilation.
- Proper uses of namespaces: You avoid the generic using namespace std which is great. 
- Adheres to the C++ standard: No issues. 
- Appropriate Modularity: I think modularity is well structured. I couldn't catch any improvement since it is very clean.
- Uses modern C++ constructs: std::optional can be used in ParseKey(return type). Also like I mentioned before constexpr can be used for "+,-,*,/".
- Uses standard library algorithms: I see usage of std::any_of and isdigit. It seems to me like you avoided extraneous loops. 
- Pointer handling: In this class, you are not using manual memory allocation which is good because there is no risk for memory leak.
- const correctness: I read the other review about this class and I agree with the suggestion on ParseKey, MakeDivideFun, MakeAddFun, MakeSubtractFun and MakeMultiplyFun.

Test:

I think you can add more test cases such as empty input {}, you can test "val1 + val2" without {}. I believe that you can add more edge cases to make sure about functionality.

Overall good job!

tranbao01:
Code Review for ReferenceVector : ( + for nice job; - for potential improvement )
1. Design:
1.1 Informative Identifiers
  + Your private member is clearly named references_, indicating it holds pointers (i.e., references) to objects of type template_type.
  + Method names such as PushBack, PopBack, Front, and Back align well with the intent of the methods.
1.2 Avoids “Magic Numbers”
  + There is no magic numbers aka hard code constants. 
1.3 Intuitive and Functional Interface
  + All member variables are private, which is good for encapsulation.
1.4 Informative Comments
  + Doxygen-style comments for each function are concise and clarify purpose.
1.5 Consistent Style
  + Indentation and bracing styles is consistent
  - Some functions like PopBack() and Back() have trailing semicolon which is not harmful but inconsistent and unusual
1.6 Visual Appearance
  + Code is clear and easy to read
2. Implementation:
2.1 No Warnings During Compilation
  +Compiles cleanly, no obvious code that triggers warnings.
2.2 Proper Uses of Namespaces
  +Everything is in namespace cse. You do not do using namespace std;. 
2.3 Adheres to the C++ Standard
  +There is no apparent undefined behavior; you’re guarding against out-of-bounds access, null pointers, etc.
2.4 Appropriate Modularity
  +The class is clear. No obvious redundant code or functionality in separate methods that could be refactored into smaller helpers.
2.5 Uses Modern C++ Constructs
  + You do use range-based for compatibility by providing begin() and end(). A const_iterator would further align with the standard library.
2.6 Uses Standard Library Algorithms Where Appropriate
  + You correctly use std::vector member functions: push_back, erase, insert, clear. No extraneous loops that could be replaced by standard algorithms.
2.7 Pointer Handling
  + You store raw pointers but do not call new or delete. This avoids memory leaks on your side.
2.8 const Correctness
  + Many methods are correctly marked const to indicate they do not modify the underlying container.
3. Testing and Error Checking
  + All methods cover in tests
  + Check both normal and error behaviors
  - Might want to consider a dedicated iterator test
Overall: 
Your code is solid. Your tests are well-written since they cover enough public methods including the edge cases

Group-05/src/DataGrid.cpp
```diff
@@ -0,0 +1,374 @@
...
+DataGrid::row_t &DataGrid::getRow(const std::size_t row_index_) {
+  assert(!grid_.empty());
+  assert(row_index_ < grid_.size());
+
+  if (grid_.empty() || row_index_ >= grid_.size()) {
+    throw std::out_of_range("Row index out of bounds");
+  }
+
+  return grid_.at(row_index_);
+}
```
tdkduong, line 49: This function is same as operator[]

Group-05/src/DataGrid.cpp
```diff
@@ -0,0 +1,374 @@
...
+  assert(column_index_ < grid_[0].size());
+  return grid_[row_index_][column_index_];
+}
+
+/**
+ * @brief Access a row in the datagrid
+ * @param row_index_
+ * @return Reference to the row at the given index
+ */
+DataGrid::row_t &DataGrid::at(std::size_t row_index_) {
```
tdkduong, line 359: These functions are same as operator[]

Group-05/src/DataGrid.cpp
```diff
@@ -0,0 +1,374 @@
...
+
+/**
+ * @brief Returns the shape of the DataGrid
+ * @return tuple of (# of rows, # of columns)
+ */
+std::tuple<const std::size_t, const std::size_t> DataGrid::shape() const {
+  size_t rows = grid_.size();
+  size_t columns = rows == 0 ? 0 : grid_[0].size();
+  return std::make_tuple(rows, columns);
+}
```
tdkduong, line 118: I don't think const in return is necessary

Group-05/src/DataGrid.cpp
```diff
@@ -0,0 +1,374 @@
...
+    row_index_ = grid_.size();
+  }
+
+  std::size_t num_columns = std::get<1>(this->shape());
+
+  // Source ChatGPT
+  // Inserts row at row_index_
+  grid_.insert(grid_.begin() + static_cast<std::ptrdiff_t>(row_index_),
+               std::vector<Datum>(num_columns, Datum(default_value_)));
+}
```
tdkduong, line 164: For this and similar function, check if grid_ is empty before inserting might be better. I'm not sure what will happen if someone insert into empty grid_.

tdkduong (commented):
DataGrid: I added some specific reviews in code for this commit,

Group-05/src/Datum.cpp
MaxKrawec: @rahnsieg Are there specific edge cases you see missing? I'm having some trouble identifying more. Thanks

tests/Group-05/DatumTests.cpp
rahnsieg, line 1: Overall Feedback:
- Adding some comments to help make the code more legible would be a good addition
- Testing setting a string datum and a double datum equal to each other
Aside from that the testing looks great!

tests/Group-05/DatumTests.cpp
```diff
@@ -0,0 +1,250 @@
...
+    CHECK_THROWS(datum_string1 / datum_double1);
+  }
+
+  SECTION("Dividing 2 string datums") {
+    cse::Datum datum_string2("test");
+    CHECK_THROWS(datum_string1 / datum_string2);
+  }
+}
+
+TEST_CASE("Equal Functionality", "[datum]") {
```
rahnsieg, line 206: Adding a test for a string datum and a double datum being compared should help ensure that all edge cases are covered.

Group-05/src/DataGrid.cpp
ShahaabAli: Thank you for the feedback. I forwarded the function to the .at member function to remove redundancy.

Group-05/src/DataGrid.cpp
ShahaabAli: I agree, thank you. I made the change

Group-05/src/DataGrid.cpp
ShahaabAli: Inserting to an empty grid works given the first assert in the function

Group-05/src/Datum.cpp
rahnsieg: I'm sorry I should have been more specific. I think I mention in DatumTests.cpp as well, but just having a test for if a user tries to compare a double datum and a string datum with each other was what I meant.

Group-05/src/Datum.h
MaxKrawec: Our group decided to leave function comments where the code was implemented. That's why some functions are "missing" comments in the header file, they are in the .cpp file.

Group-05/src/Datum.cpp
MaxKrawec: Added - Thank you

tests/Group-05/DatumTests.cpp
MaxKrawec: The comments are basically in the Section("...") part of the tests. I believe adding comments would be redundant. 

Added the test you recommended - Thanks!

tests/Group-05/DatumTests.cpp
MaxKrawec: Added the test you recommended - Thanks!

MaxKrawec:
@mercere99 and @EvilJuiceBox - This is our PR for the initial (5) classes.

Group-05/src/DataGrid.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+namespace cse {
+
+/**
+ * @brief Index a row of the DataGrid
+ * @param row_index_ Index of the row
+ * @return Indexed vector row from DataGrid
+ */
+DataGrid::row_t &DataGrid::operator[](const std::size_t row_index_) {
+  return this->at(row_index_);
```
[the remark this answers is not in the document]
mercere99: Also, such a small function should probably be fully defined in the header file.  That allows it to be inlined (it can't be inlined if the code is no in the header).

Group-05/src/DataGrid.cpp
```diff
@@ -0,0 +1,370 @@
...
+}
+
+/**
+ * @brief Index a row of the DataGrid
+ * @param row_index_ Index of the row
+ * @return Indexed vector row from DataGrid
+ */
+DataGrid::row_t &DataGrid::getRow(const std::size_t row_index_) {
+  return this->at(row_index_);
+}
```
mercere99, line 35: Again for this function (and I'm sure more below)
- Don't use `this->`
- Do a bounds check on `row_index_`
- Keep short function in the header file.

Group-05/src/DataGrid.cpp
```diff
@@ -0,0 +1,370 @@
...
+/**
+ * @brief Retrieves a row from the DataGrid in a read-only manner.
+ *
+ * Allows access to a specific row without modifying the DataGrid.
+ *
+ * @param row_index_ Index of the row to retrieve.
+ * @return A const reference to the indexed vector row from the DataGrid.
+ */
+const DataGrid::row_t &DataGrid::getRow(const std::size_t row_index_) const {
+  assert(!grid_.empty());
```
mercere99, line 46: I don't think you need to assert that the grid's not empty.  If it were, the next assert would fail.

Group-05/src/Datum.h
```diff
@@ -0,0 +1,116 @@
...
+private:
+  /// Stores the string or double value.
+  std::variant<std::string, double> value_;
+
+public:
+  /**
+   * Constructor for a string value.
+   * @param value The string value
+   */
+  explicit Datum(std::string value) { value_ = value; }
```
mercere99, line 31: Do these instructors need to be explicit?  It seems like they would work fine either way.  But consider the code:
```cpp
cse::Datum d = "test";
```
This will work if the constructor is not explicit, but (I think) will not work with the constructors set as explicit since it's an implicit conversion from `const char *` (i.e., the type of a literal string) to `std::string`.

Group-05/src/Datum.h
mercere99: This was marked as resolved, but I don't see comments here...

Group-05/src/Datum.cpp
```diff
@@ -0,0 +1,221 @@
...
+}
+
+/**
+ * Returns the double value of the Datum.
+ * @return The double value of the Datum
+ */
+double Datum::GetDouble() const {
+  if (!IsDouble()) {
+    throw std::runtime_error("Attempted GetDouble() on a non-double Datum.");
+  }
```
mercere99, line 48: Same as previous; I would probably turn this into an `assert` rather than an exception.  It's easier for a user to just check `IsDouble()` before  calling rather than building a whole try/catch structure.

Group-05/src/Datum.cpp
```diff
@@ -0,0 +1,221 @@
...
+/**
+ * Returns the string value of the Datum.
+ * @return The string value of the Datum
+ */
+std::string Datum::GetString() const {
+  if (!IsString()) {
+    // CITE: Used https://www.geeksforgeeks.org/how-to-throw-an-exception-in-cpp/
+    // to help with throwing an error.
+    throw std::runtime_error("Attempted GetString() on a non-string Datum.");
+  }
```
mercere99, line 35: I would probably turn this into an `assert` rather than an exception.  It's easier for a user to just check `IsString()` before  calling rather than building a whole try/catch structure.

Group-05/src/Datum.h
```diff
@@ -0,0 +1,116 @@
...
+   */
+  [[nodiscard]] std::variant<std::string, double> GetVariant() const { return value_; };
+
+  [[nodiscard]] std::string GetString() const;
+
+  [[nodiscard]] double GetDouble() const;
+
+  [[maybe_unused]] void AsString();
+
+  [[maybe_unused]] void AsDouble();
```
mercere99, line 89: It looks like the two functions above (`AsString` and `AsDouble`) convert the values and then don't return anything.  I would make a couple of changes here:
- The As* functions should NOT change the values, just return them As the appropriate type.  They can even be marked `const`.
- You should create To* (or maybe Set*? or Make*) versions of them that DO change the type, but even there you should return the value in the correct form.  Worst case it's not used, but since you're already doing the conversion internally, it won't mean any extra work.

Group-05/src/Datum.h
```diff
@@ -0,0 +1,116 @@
...
+
+  bool operator==(const Datum &datum) const;
+
+  /**
+   * Checks if two double or string Datums are not equal.
+   * Otherwise, returns true if one Datum is a string and the other is a double.
+   * @param datum The other Datum.
+   * @return True if the Datums are the same type and not equal each other. Otherwise false.
+   */
+  bool operator!=(const Datum &datum) const { return !(*this == datum); }
```
mercere99, line 113: You should also consider making functions `operator double()` and `operator std::string()` to allow for implicit conversions.  This will mean that you can pass a Datum variable in to a function anywhere that a string or double would have been accepted.

Group-05/src/DataGrid.cpp
```diff
@@ -0,0 +1,370 @@
...
+  assert(column_index_ < grid_[0].size());
+
+  if (grid_.empty() || column_index_ >= grid_.front().size()) {
+    throw std::out_of_range("Column index out of bounds");
+  }
+
+  ReferenceVector<Datum> column;
+
+  for (auto &row : grid_) {
+    column.PushBack(row.at(column_index_));
```
mercere99, line 122: You might want to first assert that every row is big enough, though the use of `at` should throw and exception if there's a problem.

Group-05/src/DataGrid.cpp
```diff
@@ -0,0 +1,370 @@
...
+  }
+  return column;
+}
```
mercere99, line 125: This function is exactly what I was picturing and a perfect example of how `ReferenceVector` can be really useful to have.

Group-05/src/ExpressionParser.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * @brief Evaluates equation represented by expression and returns simplified value as a double
+ *
+ * @param symbol_table_ Map with value and key that will be regerences in expression
+ * @param expression Representing equation
+ * @return double representing value of equation
+ */
+double cse::ExpressionParser::Evaluate( const std::string &expression) {
+  assert(std::any_of(expression.begin(), expression.end(), ::isdigit));
```
mercere99, line 106: Do expressions need to gave digits?  Wouldn't it be okay to have "{varA}+{varB}" as a valid expression?

Group-05/src/ExpressionParser.cpp
```diff
@@ -0,0 +1,158 @@
...
+  assert(std::any_of(expression.begin(), expression.end(), [](char c) {
+    return c=='+' || c=='-' || c=='/' || c=='*';
+  }));
```
mercere99, line 109: Similar to the above comment, I think it would be fine for an expression to be ONLY "{my_var}" without any operators.

Besides, there are so many ways that expressions can go wrong that you probably just need to check as you're processing them.

Group-05/src/ExpressionParser.cpp
mercere99: Also, while I'm normally a big fan of `asserts`, I think this might be a place where exceptions could be better?  Mostly because if it's a user-controlled expression, the programmer using your class might want to respond to a bad expression rather than just having the program crash.

Group-05/src/ExpressionParser.cpp
```diff
@@ -0,0 +1,158 @@
...
+        {
+          result = fun(symbol_table_);
+          break;
+        }
+      }
+    }
+
+  }
+  return result;
+
```
mercere99, line 156: All of the above said, for someone who hasn't taken the compilers class, this is a nice start on the functionality.

Group-05/src/ReferenceVector.h
```diff
@@ -0,0 +1,244 @@
+/**
+ * @file ReferenceVector.h
+ *
+ * @author Calen Green and Max Krawec
+ *
+ * A vector class that keeps pointers to a type of objects,
+ * allowing for entries into the vector to be treated like references.
```
mercere99, line 7: Wow -- this is a nice and very clean implementation of a ReferenceVector.  Good use of [[nodiscar]] and `const`.  Nice job having a useful iterator.

mercere99 (approved):
A set of solid classes that work together better than most other groups.  Nicely done!  Lots of comments sprinkled throughout for suggested improvements.

Group-05/src/Datum.h
MaxKrawec: @mercere99 The reviewer left another comment about the function comments later in the review. I responded to that comment. I didn't respond to this one because I thought it would be redundant, but I will in the future. If interested, below is my response.

Our group decided to leave function comments where the code was implemented. That's why some functions are "missing" comments in the header file, they are in the .cpp file.

Group-05/src/Datum.cpp
MaxKrawec: Implemented

Group-05/src/Datum.cpp
MaxKrawec: Implemented

Group-05/src/Datum.h
MaxKrawec: Implemented

Group-05/src/Datum.h
MaxKrawec: Implemented

tests/Group-05/DatumTests.cpp
```diff
@@ -0,0 +1,256 @@
...
+                 Catch::WithinAbs(987.654, kEpsilon));
+    }
+
+    SECTION("The string contains valid numbers at the start") {
+      cse::Datum string_test("123Hello");
+      string_test.AsDouble();
+      CHECK(std::isnan(string_test.GetDouble()));
+    }
+
+    SECTION("The string contains invalid numbers") {
```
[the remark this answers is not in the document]
MaxKrawec: There is a test below that does this.

tests/Group-05/DatumTests.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+  SECTION("When the Datum is a double") {
+    cse::Datum double_test(123.123);
+    double_test.AsDouble();
+    CHECK(double_test.GetDouble() == 123.123);
+  }
+}
+
+TEST_CASE("Addition functionality", "[datum]") {
+  cse::Datum datum_double1(43.26);
```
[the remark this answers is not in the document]
MaxKrawec: Fix: I refactored the tests to simplify and reduce code.

tests/Group-05/DatumTests.cpp
```diff
@@ -0,0 +1,256 @@
...
+    CHECK_THROWS(datum_double1 + datum_string1);
+  }
+
+  SECTION("Adding 1 string datum and 1 double datum") {
+    CHECK_THROWS(datum_string1 + datum_double1);
+  }
+}
+
+TEST_CASE("Subtraction functionality", "[datum]") {
+  cse::Datum datum_double1(43.26);
```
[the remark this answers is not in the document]
MaxKrawec: Implemented

tests/Group-05/DatumTests.cpp
```diff
@@ -0,0 +1,256 @@
...
+TEST_CASE("AsDouble() functionality", "[datum]") {
+  SECTION("When the Datum is a string") {
+    SECTION("The string contains all valid numbers") {
+      cse::Datum string_test("987.654");
+      string_test.AsDouble();
+      CHECK_THAT(string_test.GetDouble(),
+                 Catch::WithinAbs(987.654, kEpsilon));
+    }
+
+    SECTION("The string contains valid numbers at the start") {
```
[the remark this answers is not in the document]
MaxKrawec: Fixed

tests/Group-05/DatumTests.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+    SECTION("When the value is NaN") {
+      cse::Datum double_test(std::numeric_limits<double>::quiet_NaN());
+      double_test.AsString();
+      CHECK(double_test.GetString() == "");
+    }
+  }
+
+  SECTION("When Datum contains a string") {
+    cse::Datum string_test("hello");
```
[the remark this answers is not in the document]
MaxKrawec: Implemented

MaxKrawec (commented):
Updated Datum.cpp, Datum.h, and DatumTests.cpp based on the professor and TAs comments.

PR #136, opened 2025-02-18
Hopper

vernonriley:
Initial final


Group-03/src/Surface.cpp
```diff
@@ -0,0 +1,37 @@
+#include "Surface.h"
+
+namespace cse{
+Surface::Surface(int width, int height, int sector_size) 
+    : width(width), height(height), surface_size(sector_size), sector_width(width/sector_size), sector_height(height/sector_size){
+    
+    // I got all of this from copilot, I was going to make something similar
+    sectors.resize(sector_width, std::vector<Sector>(sector_height, Sector(0, 0))); // Resize sectors vector
```
CircleShift, line 8: I like how you're pre-allocating the inner vector, so that when the loop runs, you're not re-allocating every time you push back a new Sector!  Nice!

Group-03/src/Surface.h
```diff
@@ -0,0 +1,20 @@
...
+    public:
+        int width, height;
+        int sector_width, sector_height, surface_size;
+        std::vector<std::vector<Sector>> sectors;
+        
+        Surface(int width, int height, int sector_size);
+
+        void add_circle(Circle* circle);
+        void move_circle(Circle* circle, float x_pos, float y_pos);
+        void check_collision();
```
CircleShift, line 17: Since you will be checking for collisions, you should likely either

1.  Return the new collisions as a vector
2. Provide a way to get the collisions from the Surface in another function

My personal preference would be towards 2, since you could potentially map pointers to vectors of collisions.  In this way, a developer could `get_collisions(Circle*)` and get all the collisions on that circle

Group-03/src/Surface.cpp
```diff
@@ -0,0 +1,37 @@
+#include "Surface.h"
+
+namespace cse{
+Surface::Surface(int width, int height, int sector_size) 
+    : width(width), height(height), surface_size(sector_size), sector_width(width/sector_size), sector_height(height/sector_size){
```
CircleShift, line 5: When I compile the tests, I'm seeing

```
In file included from ../src/Surface.cpp:1:
../src/Surface.h: In constructor 'cse::Surface::Surface(int, int, int)':
../src/Surface.h:10:42: warning: 'cse::Surface::surface_size' will be initialized after [-Wreorder]
   10 |         int sector_width, sector_height, surface_size;
      |                                          ^~~~~~~~~~~~
../src/Surface.h:10:13: warning:   'int cse::Surface::sector_width' [-Wreorder]
   10 |         int sector_width, sector_height, surface_size;
      |             ^~~~~~~~~~~~
../src/Surface.cpp:4:1: warning:   when initialized here [-Wreorder]
    4 | Surface::Surface(int width, int height, int sector_size)
      | ^~~~~~~
```
Which doesn't matter too much, since they don't depend on anything not already initialized, but it does show up.

Also, since width and height are named the same in both the parameters and the class, it is a little confusing what variables they will actually reference in the sector_width and sector_height initialization

Group-03/src/Surface.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include <vector>
+#include "Sector.h"
+
+namespace cse{
+class Surface {
+    public:
+        int width, height;
+        int sector_width, sector_height, surface_size;
+        std::vector<std::vector<Sector>> sectors;
```
CircleShift, line 11: The identifiers all look informative!  It makes the code easy to understand.

Group-03/src/Surface.h
```diff
@@ -0,0 +1,20 @@
...
+        
+        Surface(int width, int height, int sector_size);
+
+        void add_circle(Circle* circle);
+        void move_circle(Circle* circle, float x_pos, float y_pos);
+        void check_collision();
+        // void update();
```
CircleShift, line 18: I'm a little worried what would happen if a circle got its position set outside the Surface, so I assume that's what this function is for, even if it's not yet implemented.  You might want to provide some way in the circle class to update whatever system it is part of, so if `set_position` is called on Circle, it calls `update` in Surface or `move_circle`

Group-03/src/Surface.cpp
```diff
@@ -0,0 +1,37 @@
...
+    }
+}
+
+void Surface::add_circle(Circle* circle) {
+    int x = circle->getX() / surface_size;   // Get the x coordinate of a circle
+    int y = circle->getY() / surface_size;   // Get the y coordinate of a circle
+    sectors[x][y].add_circle(circle);        // Add a circle to the sector
+}
+
+void Surface::move_circle(Circle* circle, float x_pos, float y_pos) {
```
CircleShift, line 22: You may want to add checks for this function (as well as `add_circle`) to make sure that the x/y of the circle is within the range of the surface.  Otherwise you may try to access a sector which is out of bounds.

Perhaps add debug asserts?

Group-03/Tests/Surface.cpp
```diff
@@ -0,0 +1,40 @@
+#include "../../third-party/Catch/single_include/catch2/catch.hpp"
+
+#include <vector>
+#include "../src/Sector.h"
+#include "../src/Surface.h"
+#include "../src/Circle.h"
+
```
CircleShift, line 7: Fine that the file is here for now, but you will likely want to move these tests to the tests folder in the top-level

Group-03/Tests/Surface.cpp
```diff
@@ -0,0 +1,40 @@
...
+    TEST_CASE("SectorTest 2", "Remove") {
+        Sector sector(0, 0);
+        Circle circle(0, 0, 1);
+        sector.add_circle(&circle);
+        sector.remove_circle(&circle);
+        CHECK(sector.circles.size() == 0);
+    }
+
+    TEST_CASE("SurfaceTest 1", "Add") {
```
CircleShift, line 24: Seems fine, you may also want to check what happens when a circle is added which is out of bounds (should probably throw an exception in debug mode at least).

Group-03/src/Surface.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
CircleShift, line 1: Overall the structure and design is fine!  Readable.

CircleShift (commented):
Review by Kyle 2/19/2025

Group-03/src/Sector.h
ShahaabAli, line 1: - Add Doxygen style comments to explain the purpose of the class variables
- Unless there is a need to make your member variables public, I'd recommend making them private (You can always use getters and setters to interact with member varriables)
- Utilize a naming convention specific to member variables to easily distinguish them form other variables
- I am not sure if there is a need for it, but I wanted to mention that if you need any special member functions, don't forget to implement them (copy/assignment constructor, move constructor, etc)

Group-03/src/Sector.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include <vector>
+#include "Circle.h"
+
+namespace cse{
+class Sector {
+    public:
+        int x, y;                       // Sector coordinates
+        std::vector<Circle*> circles;   // Circles in the sector
```
ShahaabAli, line 10: Since this is a pointer of Circle's, I would recommend making a destructor to deallocate the memory from the pointers. Though, this depends on how you are handling the pointers and if you are dynamically allocating them.

Group-03/src/Surface.h
ShahaabAli, line 1: - Add your code to the cse namespace
- Add Doxygen style comments to explain the purpose of the class and member variables
- Unless there is a need to make your member variables public, I'd recommend making them private (You can always use getters and setters to interact with member varriables)
- Utilize a naming convention specific to member variables to easily distinguish them form other variables
- I am not sure if there is a need for it, but I wanted to mention that if you need any special member functions, don't forget to implement them (copy/assignment constructor, move constructor, etc)

Group-03/src/Surface.cpp
ShahaabAli, line 1: - Add Doxygen style comments to explain the purpose of the functions and parameters
- Use asserts to test for unexpected behavior and inputs

Group-03/src/Surface.cpp
```diff
@@ -0,0 +1,37 @@
+#include "Surface.h"
+
+namespace cse{
+Surface::Surface(int width, int height, int sector_size)
```
ShahaabAli, line 4: - It would be a good idea to add asserts to check if width, height and sector_size are not negative or zero. Another check you can implement with asserts is to check if the height and width are smaller than the sector_size.
- I would recommend using size_t instead of int for your function parameters. (not 100% sure if this is better)

Group-03/src/Surface.cpp
```diff
@@ -0,0 +1,37 @@
...
+    // I got all of this from copilot, I was going to make something similar
+    sectors.resize(sector_width, std::vector<Sector>(sector_height, Sector(0, 0))); // Resize sectors vector
+    for (int i = 0; i < sector_width; ++i) {
+        for (int j = 0; j < sector_height; ++j) {
+            sectors[i][j] = Sector(i, j);       // Initialize each sector
+        }
+    }
+}
+
+void Surface::add_circle(Circle* circle) {
```
ShahaabAli, line 16: Don't forget to add some checks to see if Circle is not a nullptr, and to check if x and y are in range of your sectors vector.

Group-03/src/Surface.cpp
ShahaabAli, line 22: - Are you intentionally taking x_pos and y_pos as floats? The X and Y position member variables of surface and sector are typed as ints

Group-03/src/Surface.cpp
```diff
@@ -0,0 +1,37 @@
...
+    if (x != new_x || y != new_y) {                 // Check if the circle is in a different sector
+        sectors[x][y].remove_circle(circle);        // Remove circle from old sector
+        sectors[new_x][new_y].add_circle(circle);   // Add circle to new sector
+    }
+    circle->setPosition(x_pos, y_pos);       // Update the position of the circle
+}
+
+void Surface::check_collision() {
+    
+}
```
ShahaabAli, line 36: In case you forgot, this function is not yet implemented. This function can be made const (I am assuming this based off the name of the function.

Group-03/Tests/Surface.cpp
ShahaabAli, line 1: - These tests cases look good. There could be more tests to evaluate edge cases, which would help identify issues like out of range/bounds, negative numbers, etc.
- Also, when constructing circles and sectors, should they be constructed as raw pointers since they will be added to the surface? I am unsure why you need to construct circle and sector as an object, unless you are interacting with them outside the surface class

CircleShift, line 1: I had an error the tests:

```
../src/Sector.cpp: In member function 'void cse::Sector::remove_circle(Circle*)':
../src/Sector.cpp:13:44: error: cannot convert 'std::vector<Circle*>::iterator' to 'const char*'
   13 |     circles.erase(std::remove(circles.begin(), circles.end(), circle), circles.end());
      |                               ~~~~~~~~~~~~~^~
      |                                            |
      |                                            std::vector<Circle*>::iterator
In file included from /usr/include/c++/13.2/cstdio:42,
                 from /usr/include/c++/13.2/ext/string_conversions.h:45,
                 from /usr/include/c++/13.2/bits/basic_string.h:4097,
                 from /usr/include/c++/13.2/string:54,
                 from /usr/include/c++/13.2/stdexcept:39,
                 from ../src/Circle.h:5,
                 from ../src/Sector.h:4,
                 from ../src/Sector.cpp:1:
/usr/include/stdio.h:74:12: note:   initializing argument 1 of 'int remove(const char*)'
   74 | int remove(const char *);
      |            ^~~~~~~~~~~~
```
Which I tracked down to `Sector.h` not having the required include for `std::remove`.
I got it working by adding `#include <algorithm>` to `Sector.h`.

Group-03/src/DataTracker.cpp
```diff
@@ -0,0 +1,72 @@
...
+    if (values.empty()) return 0.0;
+    std::vector<double> sorted_values = values;
+    std::sort(sorted_values.begin(), sorted_values.end());
+    size_t size = sorted_values.size();
+    if (size % 2 == 0) {
+        return (sorted_values[size / 2 - 1] + sorted_values[size / 2]) / 2.0;
+    } else {
+        return sorted_values[size / 2];
+    }
+}
```
asifmasood7, line 44: Again, for this, I believe just having some clear documentations is needed. The functionality seems to be working well.

Group-03/src/DataTracker.cpp
```diff
@@ -0,0 +1,72 @@
...
+    double max = *std::max_element(values.begin(), values.end());
+
+    assert(std::all_of(values.begin(), values.end(), [&](double v) { return v <= max; }) && 
+           "Max value is not actually the maximum.");
+
+    return max;
+}
+
+size_t DataTracker::total() const {
+    size_t size = values.size();
```
asifmasood7, line 69: I think that since the function is simple, you might not need a local variable; directly returning values.size() would be more concise. Just saying this as a minor stylistic suggestion. :)

Group-03/src/DataTracker.cpp
```diff
@@ -0,0 +1,72 @@
...
+    
+    double min = *std::min_element(values.begin(), values.end());
+
+    assert(std::all_of(values.begin(), values.end(), [&](double v) { return v >= min; }) && 
+           "Min value is not actually the minimum.");
+
+    return min;
+}
+
+double DataTracker::max() const {
```
asifmasood7, line 57: A brief explanation here for returning 0.0 would be really helpful explaining your thought process and why you made this design choice. Or maybe also feel free to explore other alternatives like throwing an exception or using std::optional<double> could make the function clearer.

Group-03/Tests/DataTracker.cpp
```diff
@@ -0,0 +1,59 @@
+#include "../../third-party/Catch/single_include/catch2/catch.hpp"
+
+#include "../src/DataTracker.h"
+#include <vector>
```
asifmasood7, line 4: I think you can also adding tests for some additional edge cases. Some I could think of now might be:

1) Deleting a value that isn't in the tracker to ensure the function handles that part gracefully.

2) And maybe adding duplicate values and then deleting one to verify that only the first occurrence is removed.

3) Maybe some test cases with negative numbers or very large values to ensure your statistical functions like mean median handle those cases corectly.

Group-03/src/DataTracker.cpp
```diff
@@ -0,0 +1,72 @@
...
+    if (values.empty()) return 0.0;
+
+    double max = *std::max_element(values.begin(), values.end());
+
+    assert(std::all_of(values.begin(), values.end(), [&](double v) { return v <= max; }) && 
+           "Max value is not actually the maximum.");
```
asifmasood7, line 63: Just adding this comment here about the more new ideas like if you ever need to manage more complex objects instead of simple doubles, how might you use smart pointers (like std::unique_ptr or std::shared_ptr) to ensure the safe memory management in your DataTracker? Just leaving this as a future-thought!

Group-03/src/DataTracker.h
```diff
@@ -0,0 +1,21 @@
+#ifndef DATATRACKER_H
+#define DATATRACKER_H
+
+#include <vector>
+
+namespace cse {
```
asifmasood7, line 6: And just an additional question, when your group submitted the final version, were all tests passing? When I try to run everything, I get a lot of warnings and don't see some parts passing. So I wanted to make sure, keep me updated. Overall great work on your DataTracker class.

asifmasood7 (commented):
I have submitted the reviews comments for DataTracker class.

Group-03/src/FixedPoint.h
```diff
@@ -0,0 +1,71 @@
...
+    bool operator<(const FixedPoint& other) const;
+    bool operator>(const FixedPoint& other) const;
+    bool operator<=(const FixedPoint& other) const;
+    bool operator>=(const FixedPoint& other) const;
+
+private:
+    StorageType value;
+
+    static constexpr double scale() {
+        return static_cast<double>(Ratio::num) / static_cast<double>(Ratio::den);
```
MaxKrawec, line 56: Adheres to the C++ standard: Could this ever be a division by 0? May want to add an exception check.

Group-03/src/FixedPoint.cpp
```diff
@@ -0,0 +1,154 @@
...
+// Constructors
+template <typename Ratio>
+FixedPoint<Ratio>::FixedPoint() : value(0) {}
+
+template <typename Ratio>
+FixedPoint<Ratio>::FixedPoint(double d) : value(static_cast<StorageType>(d * scale())) {}
+
+template <typename Ratio>
+FixedPoint<Ratio>::FixedPoint(int64_t i) : value(i * scale()) {}
+
```
MaxKrawec, line 19: Consistent style: Used int64_t  instead of StorageType
Informative identifiers: Could potentially replace i and d with more descriptive names.

Group-03/src/FixedPoint.cpp
MaxKrawec, line 1: Informative comments: Could add comments for each function

Group-03/src/FixedPoint.cpp
```diff
@@ -0,0 +1,154 @@
...
+    fp.value = rawValue;
+    return fp;
+}
+
+// Explicit Instantiations
+template class FixedPoint<std::ratio<1>>;
+template class FixedPoint<std::ratio<10>>;
+template class FixedPoint<std::ratio<100>>;
+template class FixedPoint<std::ratio<1000>>;
+
```
MaxKrawec, line 153: Avoids “magic numbers”: Are these magic numbers? If so, may want to create constants for them.

Group-03/Tests/FixedPoint.cpp
MaxKrawec, line 1: I believe the professor wants all the test files in the 'test' folder he created. I would move my tests there.

Group-03/src/FixedPoint.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include <cstdint>
+#include <ratio>
+#include <iostream>
+#include <stdexcept>
+#include <iomanip>
+
+namespace cse {
+
+template <typename Ratio = std::ratio<1>>
```
MaxKrawec, line 12: Consistent style: Should the first letter by capitalize?

Group-03/src/FixedPoint.h
```diff
@@ -0,0 +1,71 @@
...
+class FixedPoint {
+public:
+    using StorageType = int64_t;
```
MaxKrawec, line 15: Consistent style: Should the first letter by capitalize?

Group-03/src/FixedPoint.cpp
MaxKrawec, line 1: Error handling: Could division by scale() ever by 0? If so, you should try to handle it

Group-03/Tests/FixedPoint.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+#include "../../third-party/Catch/single_include/catch2/catch.hpp"
+
+#include "../src/FixedPoint.h"
+#include <vector>
+#include <sstream>
+
+using Ratio1 = std::ratio<100>;
+using Fixed1 = cse::FixedPoint<Ratio1>;
+
```
MaxKrawec, line 11: Consistent style: Should the first letters by capitalize for these?

MaxKrawec (commented):
Design:
1) Informative identifiers [Minor Suggestions] - Overall, good job. I would update some of the variable names to be more descriptive.
2) Avoids “magic numbers” [Minor Suggestion] - Potentially uses magic numbers for the Explicit Instantiations.
3) Intuitive and functional interface [No Suggestions] - Very good job. Member variable and helper functions are private, while the accessible functions are public.
4) Informative comments [Suggestion] - I would add comments to your functions.
5) Consistent style [Minor Suggestions] - I would talk to your team about using #pragma once. Also, you used int64_t instead of StorageType sometimes. Also, make sure to keep the same style for naming variables. 
6) Visual appearance: [No Suggestions] - Excellent use of spaces and white space.

Implementation:
1) No warnings during compilation [Major Problem] - I tried to run your tests, but couldn't due to an error. Although I believe your class had no warnings, your group should investigate the source of the error.
2) Proper uses of namespaces: [No Suggestions] - All code is written within the namespace cse and doesn't have pollution.
3) Adheres to the C++ standard [Major Problem] - As mentioned above, I couldn't compile your code due to an error. Your group should investigate the source of the error. Also, could the scale function ever have a division by 0?
4) Appropriate Modularity [No Suggestions] - Good job creating helper functions to reduce duplicate or unnecessary code.
5) Uses modern C++ constructs [No Suggestions] - Good job using std::ratio which is relatively new.
6) Uses standard library algorithms [No Suggestions] - Doesn't need the standard library algorithms
7) Pointer handling: [No Suggestions] - Doesn't use pointers.
const correctness [No Suggestions] - Excellent use of const throughout.

Testing and Error Checking:
1) Unit testing [No Suggestions] - Unit tests are present for all public member functions and are well organized. Also tests the private functions
2) Edge cases [Suggestions] - Good job checking for division by 0. These are ChatGPT suggestions: 1) Check for overflow 2) Handling of extreme ratios
3) Assertions [Suggestions] - Doesn’t use asserts. Could add some to check that valid arguments are passed into the functions. 
4) Error handling [Suggestions] Could implement the checks recommended above: 1) Check for overflow 2) Handling of extreme ratios

KrishPatel198:
### **Review for Circle**


**Design:**

Informative Identifiers: Variable and function names are clear and descriptive. No unnecessary abbreviations.
Avoids Magic Numbers: No arbitrary constants are hardcoded. The class properly validates the radius but could define a constexpr double MIN_RADIUS = 0.0 for consistency.
Intuitive and Functional Interface: Well-structured with private members and appropriate getters/setters. Constructor ensures valid initialization.
Informative Comments: Lacks inline documentation. Adding Doxygen-style comments would improve readability.
Consistent Style: Uses a consistent naming convention (camelCase for methods and snake_case for variables). Follows standard header guards and includes.
Visual Appearance: Code is well-indented and structured. Proper spacing and formatting make it easy to read.


**Implementation:**

No Warnings During Compilation: No apparent warnings or errors. Proper use of #pragma once and exception handling.
Proper Uses of Namespaces: No namespace pollution. Does not use using namespace std;, which is good practice.
Adheres to C++ Standard: Uses standard library (cmath, stdexcept) and follows best practices.
Appropriate Modularity: The class is self-contained and adheres to SRP. No unnecessary duplication.
Uses Modern C++ Constructs: Uses exception handling and const correctness. Could benefit from std::optional for characteristic_ instead of an empty string.
Uses Standard Library Algorithms: Uses std::sqrt, avoiding unnecessary loops.
Pointer Handling: No manual memory management required.
Const Correctness: Proper use of const for getters and overlaps(). The setCharacteristic function could accept const std::string& to avoid unnecessary copying.


**Testing and Error Checking:**

Unit Testing: Includes multiple test cases using Catch2, ensuring constructor validity, invalid radius handling, and overlap behavior.
Edge Cases: Tests invalid radii but should also check circles with zero radius and extremely large values.
Assertions: Uses exceptions for error handling but does not assert invalid states.
Error Handling: Ensures radius is positive but could explicitly handle NaN values for robustness.


**Final Thoughts:**

This is a well-implemented class with strong error handling and clear structure. Improvements include adding documentation, using constexpr for constants, using std::optional for characteristic_, and passing const std::string& in setters to optimize performance.

Group-03/src/FunctionSet.cpp
```diff
@@ -0,0 +1,116 @@
...
+                const PointerType* storedPtr = storedFunc.template target<PointerType>();
+                if (storedPtr && *storedPtr == *funcPtr)
+                {
+                    return i;
+                }
+            }
+        }
+    }
+    return -1;
+}
```
gringrant, line 98: Idea: Instead of using a special int value to represent the state where there is no index, perhaps use a std::optional to represent the idea that there might not be a meaningful value returned.

Group-03/src/FunctionSet.cpp
```diff
@@ -0,0 +1,116 @@
...
+ */
+template <typename R, typename... Args>
+void cse::FunctionSet<R, Args...>::removeFunction( const FunctionType& func)
+{
+    int index = findFunctionIndex(func);
+    if (index >= 0)
+    {
+        mFunctions.erase(mFunctions.begin() + index);
+    }
+}
```
gringrant, line 113: Idea: a remove function should report if it successfully removed anything or if it couldn't find anything. (usually through a bool return)

gringrant (commented):
Very good work! Most of these comments are about behaviors I would expect some common functions to have.

chrispaul-lab:
**Review for FixedPoint**

_**Design**_

_Informative Identifiers:_
1. Names are clear and well-chosen, effectively communicating purpose. No improvements needed.

_Avoids “magic numbers”_
1. The scale() function properly abstracts fixed-point scaling, ensuring maintainability. No issues.

_Intuitive and functional interface_
1. Interface is well-designed, exposing only necessary operations while keeping implementation details private. No changes needed.

_Informative comments_
1. More detailed explanations of the fixed-point representation would improve clarity.

_Consistent style_
1. Naming conventions follow a consistent pattern. No inconsistencies.

_Visual appearance_
1. Code is well-structured. No issues.

_**Implementation**_
1. No warnings during compilation
2. Proper uses of namespaces
3. Adheres to the C++ standard
4. Uses modern C++ constructs
5. Arithmetic operations are directly implemented, so STL use not necessary right now. 
6. No raw pointers
7. const correctness taken care of. 

_**Test Cases & Error Checking**_
_Assertions:_
1. More assertions could validate inputs or internal states where appropriate.

Everything else looks good.

Group-03/Tests/DataTracker.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+
+TEST_CASE("DataTrackerTest 1", "Empty") {
```
132ikl, line 7: Style: You should provide a description as the first argument to `TEST_CASE`, and the second argument should be in brackets and be the same for all tests in your file. For example:

```suggestion
TEST_CASE("DataTracker empty", "[DataTracker]") {
```

132ikl (commented):
I reviewed `DataTracker`. I expanded upon this in another comment, but you should consider a different internal representation for your data which is more efficient for the specific operations you're doing (median, min, max). Some of your tests don't compile (which halts compilation of the remaining tests), but I was able to verify that the tests for `DataTracker` specifically compiled without warnings and passed.

There's also a couple places where you could make local variables `const`, for example, each time you call `size()`, and when you compute `min` and `max`.

Overall, looks pretty good :smile:

Group-03/src/DataTracker.h
132ikl: I also have this issue, it seems like there's an issue with the Sector class. If I remove the other classes, the `DataTracker` tests seem to pass.

Group-03/src/Circle.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+void Circle::setRadius(double radius) {
+    if (radius <= 0) {
+        throw std::invalid_argument("Radius must be positive");
+    }
+    radius_ = radius;
+}
+
+// Check if this circle overlaps with another circle
+bool Circle::overlaps(const Circle& other) const {
```
PMitkiewicz, line 41: If someone has multiple circles they want to compare them too, which I believe would be a common use case, you would need to call this once for every circle squared. Instead, I recommend having the function take a container of circles, so you could just compare it with all the circles as once, which makes calls easier. Even better, have the function simply compare every circle to every other circle within a container, so you would only need one call. If you still want to be able to compare each circle individually, you can just make a container with only those two circles. Just an idea though, let me know if you'd like to discuss this further.

Group-03/src/Circle.cpp
```diff
@@ -0,0 +1,55 @@
+// Circle.cpp : implementation of the Circle class
+#include "Circle.h"
+
+// Constructor
+Circle::Circle(double x, double y, double radius) : x_(x), y_(y), radius_(radius) {
```
PMitkiewicz, line 5: x and y are not very informative names, it took a little bit to realize its the position of the center of the circle. I think something like center_x_ and center_y_ would be better.

Group-03/src/Circle.cpp
```diff
@@ -0,0 +1,55 @@
...
+// Check if this circle overlaps with another circle
+bool Circle::overlaps(const Circle& other) const {
+    double dx = x_ - other.x_;
+    double dy = y_ - other.y_;
+    double distance = std::sqrt(dx * dx + dy * dy);
+    return distance < (radius_ + other.radius_);
+}
+
+// Miscellaneous characteristics
+void Circle::setCharacteristic(const std::string& characteristic) {
```
PMitkiewicz, line 49: I'm a little confused on what these miscellaneous characteristics would be. Perhaps better documentation would help clarify things.

Group-03/src/Circle.h
PMitkiewicz, line 1: I recommend doc strings, or at least more comprehensive comments. For example, I have a hard time telling what exactly characteristics are or what they are used for.

Group-03/src/Circle.cpp
PMitkiewicz, line 1: Overall, very good job, it's well written code and it performs the base functionality well. There are some changes I think you could make for usability , as I commented on the overlaps function, and some better naming and documentation. Other than that user side stuff, it looks great.

PMitkiewicz: Side note, style is consistent and it's easy to read through. Biggest area of improvement is definitely the comments.

Group-03/src/FunctionSet.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include "FunctionSet.h"
+
+#include <iostream>
+#include <string>
+
+
+/**
+ * Constructor
+ */
+template <typename R, typename... Args>
```
greencalen17, line 15: I believe we want all template implementations in the header file not the cpp file.

Group-03/src/FunctionSet.cpp
```diff
@@ -0,0 +1,116 @@
...
+bool cse::FunctionSet<R, Args...>::isEmpty() const
+{
+    return mFunctions.empty();
+}
+
+/**
+ * countFunSet
+ */
+template <typename R, typename... Args>
+std::size_t cse::FunctionSet<R, Args...>::countFunSet() const
```
greencalen17, line 66: I feel like size or count is a better name for this function.

Group-03/src/FunctionSet.cpp
```diff
@@ -0,0 +1,116 @@
...
+ * Destructor
+ */
+template <typename R, typename... Args>
+cse::FunctionSet<R, Args...>::~FunctionSet() = default;
+
+/**
+ * addFunction
+ */
+template <typename R, typename... Args>
+void cse::FunctionSet<R, Args...>::addFunction(const FunctionType& func)
```
greencalen17, line 28: For this function and all others, there should be documentation for the parameters and the return values where applicable.

Group-03/src/FunctionSet.cpp
```diff
@@ -0,0 +1,116 @@
...
+void cse::FunctionSet<R, Args...>::clearAll()
+{
+    mFunctions.clear();
+}
+
+/**
+ * isEmpty
+ */
+template <typename R, typename... Args>
+bool cse::FunctionSet<R, Args...>::isEmpty() const
```
greencalen17, line 57: could mark with [[nodiscard]] to make sure the value is used.

Group-03/src/FunctionSet.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * addFunction
+ */
+template <typename R, typename... Args>
+void cse::FunctionSet<R, Args...>::addFunction(const FunctionType& func)
+{
+    using PointerType = R(*)(Args...);
+
+    if (func.target_type() == typeid(PointerType)) {
```
greencalen17, line 32: may want to check that func is !null first.

Group-03/src/FunctionSet.cpp
Thisisnotgoingtowork: Weneed to have this so that this work in from header to here, not you want or not...

Group-03/src/DataTracker.cpp
```diff
@@ -0,0 +1,72 @@
...
+namespace cse {
+void DataTracker::add_value(double value) {
+    size_t old_size = values.size();
+    values.push_back(value);
+
+    assert(values.size() >= old_size && "Size did not increase");
+
+}
+
+void DataTracker::delete_value(double value) {
```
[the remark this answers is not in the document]
nigamsa1: That is a great point and I have added some comments throughout!

Group-03/src/DataTracker.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    assert(mean >= min() && mean <= max() && "Mean should be within the range.");
+
+    return mean;
+}
+
+double DataTracker::median() const {
+    if (values.empty()) return 0.0;
+    std::vector<double> sorted_values = values;
+    std::sort(sorted_values.begin(), sorted_values.end());
```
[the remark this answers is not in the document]
nigamsa1: I think this is something we intend to improve on in our advanced class!

Group-03/src/DataTracker.h
nigamsa1: That is something someone else in our group has been struggling with so it is not just you! We are still working on that

Group-03/src/DataTracker.h
```diff
@@ -0,0 +1,21 @@
...
+class DataTracker {
+private:
+    std::vector<double> values; //main container
```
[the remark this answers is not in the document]
nigamsa1: Will add in our advanced class

Group-03/src/Surface.h
mercere99: Depending how you handle this function, you can probably also make it `const`.

Group-03/src/Circle.cpp
```diff
@@ -0,0 +1,55 @@
+// Circle.cpp : implementation of the Circle class
+#include "Circle.h"
+
+// Constructor
+Circle::Circle(double x, double y, double radius) : x_(x), y_(y), radius_(radius) {
+    if (radius <= 0) {
+        throw std::invalid_argument("Radius must be positive");
+    }
```
mercere99, line 8: I would make this an `assert` rather than a `throw` -- it should never be called with a non-positive radius.

Group-03/src/Circle.cpp
```diff
@@ -0,0 +1,55 @@
...
+// Setters
+void Circle::setPosition(double x, double y) {
+    x_ = x;
+    y_ = y;
+}
+
+void Circle::setRadius(double radius) {
+    if (radius <= 0) {
+        throw std::invalid_argument("Radius must be positive");
+    }
```
mercere99, line 36: Likewise, this should be an `assert`.

Group-03/src/Surface.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+void Surface::add_circle(Circle* circle) {
+    int x = circle->getX() / surface_size;   // Get the x coordinate of a circle
+    int y = circle->getY() / surface_size;   // Get the y coordinate of a circle
+    sectors[x][y].add_circle(circle);        // Add a circle to the sector
+}
+
+void Surface::move_circle(Circle* circle, int x_pos, int y_pos) {
+    int x = circle->getX() / surface_size;          // Get the x coordinate of a circle
```
mercere99, line 23: As @ShahaabAli points out above, you should also add asserts to this function to make sure the `Circle` being added is valid for the `Surface`.

mercere99 (approved):
Lots of fixes needed (pointed out throughout the code and peer reviews), but can clean up to hook this code into main.

PR #137, opened 2025-02-18
Group 4 Classes

132ikl:
This PR contains the most up-to-date versions of all group 4 classes for review.


KrishPatel198:
2648d68

CircleShift:
Any commit b580226f847c04d38a5633c10a76bba8d6c02d11 or later should be fine to test BitVector on

Group-04/lib/RichText.hpp
ShahaabAli, line 1: Overall feedback:
- Use consistent naming scheme for member variables and member functions to clear indicate them as member variables/functions. It may also be worth considering renaming some variables and functions to more clearly indicate their purpose.
- Include doxygen style comments for classes, member variables, functions and standard comments for more clarity because the design and purpose of some parts of your code is unclear
- Add asserts to verify expected behavior, especially for out of bounds checking
- Consider using a .cpp for lengthier functions, unless there is a specific reason for only using an .hpp

Group-04/lib/RichText.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+  const char& char_at(size_t pos) const { return text.at(pos); }
+
+  std::unique_ptr<std::vector<Format>> formats_at(size_t pos) const {
+    auto result = std::make_unique<std::vector<Format>>();
+    for (const auto& [format, index] : formatting) {
+      if (index.contains(pos)) result->push_back(format);
+    }
+    return result;
+  }
```
ShahaabAli, line 66: An assert could be used to check if pos is in bounds

Group-04/lib/RichText.hpp
```diff
@@ -0,0 +1,110 @@
...
+    return *this;
+  }
+
+  void applyFormatToRange(const Format& format, const size_t begin,
+                          const size_t end) {
+    auto [item, inserted] = formatting.insert({format, IndexSet(begin, end)});
+    if (!inserted) {
+      item->second.insertRange(begin, end);
+    }
+  }
```
ShahaabAli, line 99: Assert that begin and end are in valid bounds

Group-04/lib/RichText.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+  RichText& operator+=(const RichText& str) { return append(str); }
+
+  std::optional<IndexSet> getFormatRange(const Format& format) const {
+    auto iter = formatting.find(format);
+    if (iter == formatting.end()) return {};
+    return {iter->second};
+  }
```
ShahaabAli, line 107: Assert if format is valid

Group-04/lib/RichText.hpp
ShahaabAli: I do not see any tests that evaluate the functionality of this class for both base and edge cases. Make sure to write some to check for bugs and better garuntee functionality.

matrix373 (commented):
Review for RichText class

Group-04/lib/BitVector.hpp
```diff
@@ -0,0 +1,854 @@
...
+      3, 4, 4, 5, 4, 5, 5, 6, 4, 5, 5, 6, 5, 6, 6, 7,  // clang-format
+      2, 3, 3, 4, 3, 4, 4, 5, 3, 4, 4, 5, 4, 5, 5, 6,  // won't
+      3, 4, 4, 5, 4, 5, 5, 6, 4, 5, 5, 6, 5, 6, 6, 7,  // destroy
+      3, 4, 4, 5, 4, 5, 5, 6, 4, 5, 5, 6, 5, 6, 6, 7,  // my nice rectangle.
+      4, 5, 5, 6, 5, 6, 6, 7, 5, 6, 6, 7, 6, 7, 7, 8,  // 0b11111111
+  };
+
+  // Keep the bottom "keep" bits of the given byte, discarding the rest and
+  // updating the num_set
+  void fixup_byte(size_t byte, uint8_t keep);
```
rahnsieg, line 46: I would suggest changing the name of this function to something more descriptive.

Group-04/lib/BitVector.hpp
```diff
@@ -0,0 +1,854 @@
...
+  if (num_bits <= idx)
+    throw std::out_of_range(std::format(
+        "Invalid index into BitVector: idx - {}, max - {}", idx, num_bits));
+#endif
+  const std::byte b = (std::byte{1} << (idx % 8)) & underlying[idx / 8];
+  return b != std::byte{0};
+}
+
+// Set bits in a pattern
+BitVector& BitVector::pattern_set(size_t start, size_t count,
```
rahnsieg, line 301: I would suggest adding some additional comments in the function that explain how the code works

Group-04/lib/BitVector.hpp
```diff
@@ -0,0 +1,854 @@
...
+// Left-shift and set operator
+BitVector& BitVector::operator<<=(size_t pos) {
+  if (pos == 0)
+    return *this;
+  else if (pos >= num_bits) {
+    reset();
+    return *this;
+  }
+
+  num_set = 0;
```
rahnsieg, line 656: I would suggest having a specific naming convention for private member variables

Group-04/lib/BitVector.hpp
rahnsieg, line 1: Overall Feedback:
- Make sure you have asserts for all public member functions in your class
- I suggest having a specific naming convention for member variables
- Make sure all the comments in the class are informative and meaningful
- Make sure there are tests for all the public member functions
Overall the class looks really good and it is very clear that you know what you are doing!

Group-04/bin/CseStringTest.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+#include <cassert>
+#include <iostream>
+
+#include "CseString.hpp" 
+
+// Helper function to print test results
+inline void print_test_result(const char* test_name, bool passed) {
+  std::cout << test_name << ": " << (passed ? "PASSED" : "FAILED") << std::endl;
+}
```
IvanBega, line 20: In addition to printing output to the console, you can also redirect the output into a stringstream and compare it directly in your code

Group-04/lib/CseString.hpp
```diff
@@ -0,0 +1,67 @@
...
+    dbg_assert(pos <= size(), "cse::String substr() start position out of range");
+    // Use base-class std::string::substr, then wrap the result in cse::String
+    return cse::String(std::string::substr(pos, count));
+  }
+};
+
+#else
+
+// In Release builds, String is just an alias to std::string
+using String = std::string;
```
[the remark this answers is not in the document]
IvanBega: I have also reviewed tests located in /bin/CseStringTest.hpp. They appear to cover all functionality from the audited string. Each function has extensive Doxygen documentation, and every test inside functions also has comments that describes it. Tests look solid and account for edge cases, most negative test that trigger assert error had to be commented so they don't crash the program.

I couldn't find catch2 tests in /tests/Group-04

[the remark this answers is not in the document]
IvanBega: I compiled CseString.hpp with CseStringTest.hpp separately using g++ 23. -Wall -Wextra and didn't have any warnings. However, I couldn't compile whole solution with the makefile because I didn't have #include <format> on my machine, even though I have g++, c++23 installed

Group-04/lib/BitVector.hpp
MaxKrawec, line 1: Sometimes you have `#ifndef NDEBUG` and `#endif` around your exception checks, and other times you don't. Is this intentional?

tests/Group-04/BVOps.cpp
```diff
@@ -0,0 +1,310 @@
...
+  CHECK(bv2 == bc2);
+
+  // Pattern set part of byte
+  cse::BitVector bv4(8), bc4("00011000");
+  bv4.pattern_set(2, 3, std::byte{0b01010110});
+
+  CHECK(bv4.count() == bc4.count());
+  CHECK(bv4 == bc4);
+
+  // Pattern set part of byte (offset)
```
MaxKrawec, line 99: Are you missing tests here?

MaxKrawec (commented):
Design
1) Informative identifiers [Minor Suggestions]: 
1.1) Throughout your code, you sometimes use 1 or 2 letters for variable names. Although it’s still readable, I personally prefer more descriptive names.
2) Avoids “magic numbers” [Minor Suggestions]
2.1) You use 8 a lot throughout your code. Personally, I would create a constant for it. For example, static constexpr size_t BITS_PER_BYTE = 8
2.2) You also use 0b11111111 in some of your functions. Could make that a constant as well
3) Intuitive and functional interface [No Suggestions]
3.1) I noticed you copied many function names from std::bitset. That was a smart idea because if someone is familiar with std::bitset, they should be able to use your class easily.
4) Informative comments [Minor Suggestions]
4.1) Personally, I like using doxygen comments which provides information on the params, returns, warnings, etc. However, you did a great job providing inline comments to your code. 
5) Consistent style [Minor Suggestions]
5.1) Sometimes your variable names are abbreviated and sometimes they aren’t. For example, you abbreviate index to idx, but not count to cnt. When do you choose to abbreviate and when not to? Personally, I try not to abbreviate variable names. 
5.2) Also, the spacing between tests seems a little off. Some have 4 spaces while others don’t.
6) Visual Appearance [No Suggestions]
6.1) It looks like you use a clang format, so the code looks nice. Good job!

Implementation
1) No warnings during compilation [No suggestions]
1.1) I received no warnings and all the tests passed. Nice!
2) Proper uses of namespaces [No Suggestions]
2.1) You have a namespace cse and you don’t have namespace pollution.
3)Adheres to the C++ standard [No Suggestions]
3.1) You have exception checks throughout to look for undefined behavior. Good job.
4) Appropriate Modularity [Minor Suggestion]
4.1) The pattern_set function is a bit long and confusing for me. Although you added comments throughout which helped, I would try to break it up into a few helper functions. I believe this would increase readability.  
5) Uses modern C++ constructs [Minor Suggestion]
5.1) I believe you could use a for each loop in some places.
5.1.1) For example, line 428.
5.2) Uses std::byte and std::format which are relatively modern C++ techniques (C++17+).
6) Uses Standard Library Algorithms [Minor Suggestion] - Note: Suggestions from Claude
6.1) Some loops could potentially use std::fill or std::copy
6.2) Could use std::count for counting set bits
7) Pointer handling [No Suggestions]
7.1) Not needed
8) const correctness [No Suggestions]
8.1) Many functions are const and have const parameters. Good job!

Testing and Error Checking
1) Unit Testing [Minor Suggestions]
1.1) Is there a test for the any and none functions? I couldn’t seem to find them.
1.2) Also, it may be unnecessary to have a .cpp file for size tests. Since it’s only 2 tests, I would add them to the other test file. 
2) Edge Cases [No Suggestions]
2.1) Catches out of range errors and empty values.
3) Assertions [Suggestion]
3.1) Maybe I’m missing them, but I don’t see any asserts or custom asserts. If true, I recommend adding asserts for ensuring a valid state. 
4) Error handling [No Suggestion]
4.1) You have multiple error checks and you do a good job handling them.

aydinorh:
Review for Assert:
Design:

- Informative identifiers: In line 139 instead of name 'a', you can use more descriptive name. Other than that, types and names are all good and understandable.
- Avoids “magic numbers”: I didn't see any magic numbers being used. 
- Intuitive and functional interface: You can make these variables private and use getter functions to retrieve them.
        bool condition;
        std::optional<const char *> message = std::nullopt;
- Informative comments: I would use Doxygen documentation for all of the functions. It would be good for a better readability.
- Consistent style: For your member variable "condition" and "message" you can use m_condition, m_message if you decide to make them private.
- Visual appearance: Overall, your code visually appears very good. 

Implementation:

- No warnings during compilation: I checked your code and there was no warning during compilation.
- Proper uses of namespaces: You avoid the generic using namespace std which is very good. 
- Adheres to the C++ standard: No issues. 
- Appropriate Modularity: I think your code is fairly modular. I couldn't catch any improvement.
- Uses modern C++ constructs: You are using std::optional, constexpr and requires clauses. 
- Uses standard library algorithms: You already avoid extraneous loops and barely need standard library algorithms.
- Pointer handling: You are not using manual memory allocation. 
- const correctness: PrintAssertMessage and PrintLocation functions can be declared as const because they are not modifying the original class.

Test:

Your edge cases should be tested such as empty containers in your "Test assert equal". Also you should check for handling unexpected input types to catch misuses. Overall test looking solid.

Group-04/lib/BitVector.hpp
CircleShift: Ultimately I agree, I think I will change this to "truncate_byte" since "fixup" could probably mean anything

Group-04/lib/BitVector.hpp
CircleShift: Noted, since most groups are familiar with the format, I will change to the `m_` prefix format

Group-04/lib/BitVector.hpp
CircleShift: Good feedback, will look into this

Group-04/lib/BitVector.hpp
CircleShift: Yep!  It's supposed to mirror the behavior of e.g. vector or bitset where member functions will throw exceptions while indexing or "primitive" operations may result in UB.  I've tried to make it so that UB will only occur in release mode so debug mode may be safer.

Any thoughts on this?

Group-04/lib/BitVector.hpp
```diff
@@ -0,0 +1,854 @@
...
+
+  // Keep the bottom "keep" bits of the given byte, discarding the rest and
+  // updating the num_set
+  void fixup_byte(size_t byte, uint8_t keep);
+
+ public:
+  // Author(s): Kyle Gunger
+  // BitVector::reference is a reference to a single bit in the vector,
+  // providing common bitwise operations as well as decomposing into a boolean
+  class reference {
```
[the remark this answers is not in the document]
CircleShift: I was trying to mirror `std::bitset::reference`, so I used the same naming style

Group-04/lib/BitVector.hpp
```diff
@@ -0,0 +1,854 @@
...
+  *this = bv;
+  append(tmp);
+}
+
+BitVector& BitVector::operator+=(const BitVector& bv) {
+  this->append(bv);
+  return *this;
+}
+
+BitVector BitVector::operator+(const BitVector& bv) const {
```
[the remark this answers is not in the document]
CircleShift: Noted!  Thanks for pointing that out

tests/Group-04/BVOps.cpp
CircleShift: Oh yeah probably

tests/Group-04/Assert.cpp
```diff
@@ -0,0 +1,78 @@
...
+TEST_CASE("Test basic assert functionality", "[assert]") {
+  // test most basic asserts
+  REQUIRE_NOASSERT(cse_assert(true));
+  REQUIRE_ASSERT(cse_assert(false));
+
+  // test assert with message
+  REQUIRE_NOASSERT(cse_assert(true, "hello world"));
+  REQUIRE_ASSERT(cse_assert(false, "hello world"));
+
+  // test asserts with comma in condition
```
asifmasood7, line 20: I was thinking that maybe you can consider adding tests for some more complex scenarios. Maybe you can include tests that involve complex expressions or side effects in the condition to ensure single evaluation. Could be scenarios where the condition involves more than one comma (beyond just std::is_same_v usage) to further validate your variadic macro splitting.

tests/Group-04/Assert.cpp
```diff
@@ -0,0 +1,78 @@
...
+  // ensure all are no-ops when debug mode is disabled
+  REQUIRE_NOASSERT(dbg_assert(true));
+  REQUIRE_NOASSERT(dbg_assert(false));
+  REQUIRE_NOASSERT(dbg_assert(true, "hello world"));
+  REQUIRE_NOASSERT(dbg_assert(false, "hello world"));
+  REQUIRE_NOASSERT(cse_assert_eq(1, 2));
+  REQUIRE_NOASSERT(cse_assert_eq(1, 1));
+  REQUIRE_NOASSERT(cse_assert_neq(1, 1));
+  REQUIRE_NOASSERT(cse_assert_neq(1, 2));
+#endif
```
asifmasood7, line 77: Completely optional - but just a thought. Even though it can be challenging to capture console output in unit tests, consider if there’s a way to verify that the error messages include the expected file, line, and function details. This might involve redirecting output streams or maybe thru using a logging interface.

tests/Group-04/Assert.cpp
```diff
@@ -0,0 +1,78 @@
...
+  REQUIRE_ASSERT(cse_assert_eq(1, 2));
+  REQUIRE_NOASSERT(cse_assert_eq(1, 1));
+
+  // basic assert_eq (non-printable)
+  REQUIRE_ASSERT(cse_assert_eq(std::vector<int>{0}, std::vector<int>{1}));
+  REQUIRE_NOASSERT(cse_assert_eq(std::vector<int>{0}, std::vector<int>{0}));
+}
+
+TEST_CASE("Test assert not equal", "[assert]") {
+  // basic assert_eq (printable)
```
asifmasood7, line 43: Whenever you have a moment, I think that adding a brief comment at the top of the test file or before each section summarizing what is being tested can help more in detail to understand the scope of each test case quickly.

tests/Group-04/Assert.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+TEST_CASE("Test assert never", "[assert]") {
+  REQUIRE_ASSERT(cse_assert_never());
+  REQUIRE_ASSERT(cse_assert_never("hello world"));
+}
+
+TEST_CASE("Test assert equal", "[assert]") {
+  // basic assert_eq (printable)
+  REQUIRE_ASSERT(cse_assert_eq(1, 2));
```
asifmasood7, line 34: Another side note that you can try to ensure that you test both variants of each macro (with and without messages) consistently. You’re doing that in many cases; just double-check if there’s any assert type that might benefit from an extra test case. For instance, you might consider an additional test case for cse_assert_eq or cse_assert_neq with custom types or edge-case values, if applicable.

asifmasood7 (commented):
Reviewed the Assert class and left comments. Great work, Rose!

Group-04/lib/BitVector.hpp
rahnsieg: I would stick with how mirroring the bahavior of std::vector to the best of your abilities.

MaxKrawec: I believe you talked about this with the professor during your presentation. It sounds like he likes your current implementation, so I would keep it as is. Great job! 

Side note: A lot of my comments are in the 'comment' for this review. Let me know if you can't find it.

Group-04/bin/BVTest.hpp
rahnsieg, line 1: Overall Feedback:
- I can't seem to find any tests for a bitvector of size 0 in the file and I would recommend adding one
- I would also recommend adding a test for the max size of bitvector
Aside from that everything looks great!

Group-04/lib/IndexSet.hpp
```diff
@@ -0,0 +1,574 @@
...
+    std::size_t range_idx_;
+    std::size_t current_;
+  };
+
+  // Constructors
+  IndexSet() = default;
+  ~IndexSet() = default;
+  IndexSet(size_t start, size_t end) {
+    ranges_.emplace_back(start, end);
+    total_size_ = end - start;  // Fix: Set correct total size
```
greencalen17, line 156: If this has already been fixed, remove the comment.

Group-04/lib/IndexSet.hpp
```diff
@@ -0,0 +1,574 @@
...
+  }
+
+  /**
+   * @brief Removes an index from the set
+   * @param index The index to remove
+   *
+   * If the index is in the middle of a range, the range will be split.
+   * For example, if [1,5) exists and we remove 3, we get [1,3) and [4,5).
+   */
+  void remove(std::size_t index) {
```
greencalen17, line 249: Same here, I think index can be marked const

Group-04/lib/IndexSet.hpp
```diff
@@ -0,0 +1,574 @@
...
+
+    // Verify the range actually contains the index
+    if (range_idx < ranges_.size() && ranges_[range_idx].first <= index &&
+        ranges_[range_idx].second > index) {
+      return ranges_[range_idx];
+    }
+    return std::nullopt;
+  }
+
+  void offset(const size_t offset) {
```
greencalen17, line 381: Document this function

PraiseHeavyLobster:
Index Set:
Design:
-Informative Identifiers: Types and names are good and understandable
-Avoids "Magic Numbers": No magic numbers are used.
-Intuitive and functional interface: interface is intuitive and there are no public member variables.
-Informative comments: Comments are informative. I would recommend using Doxygen for the member variables at lines 515 and 516
-Consistent Style: Function naming is a little inconsistent, with some functions such as merge_overlapping_ranges (line 526) using snake_case, and getContainingRange (line 369) using camelCase 
-Visual Appearance: code is easy to scan through and visually well structured. Code structure is consistent

Implementation:
-No warnings during compilation: There are no warnings in compilation
-Proper use of namespaces: all code is within namespace cse, and the code avoids using the generic using namespace std as well as avoiding namespace pollution
-Adheres to the C++ standard: Code adheres to C++ standard and avoids undefined behavior
-Appropriate Modularity: Code is grouped into appropriate modularity
-Uses modern C++ constructs: Code uses for loops and std::optional
-Uses standard library algorithms: no issue
-Pointer Handling: pointer uses raw pointers, but without new and delete. No use of either smart pointers std::unique_ptr and std::shared_ptr
-const correctness: no issue

Test:
-There is no testing for getNextRange, getPrevRange, and getContainingRange. Other than that, testing looks good and comprehensive!

Group-04/bin/CseStringTest.hpp
PMitkiewicz: I agree this would be a nice addition

PraiseHeavyLobster:
Index Set:
Design:
-Informative Identifiers: Types and names are good and understandable
-Avoids "Magic Numbers": No magic numbers are used.
-Intuitive and functional interface: interface is intuitive and there are no public member variables.
-Informative comments: Comments are informative. I would recommend using Doxygen for the member variables at lines 515 and 516
-Consistent Style: Function naming is a little inconsistent, with some functions such as merge_overlapping_ranges (line 526) using snake_case, and getContainingRange (line 369) using camelCase
-Visual Appearance: code is easy to scan through and visually well structured. Code structure is consistent

Implementation:
-No warnings during compilation: There are no warnings in compilation
-Proper use of namespaces: all code is within namespace cse, and the code avoids using the generic using namespace std as well as avoiding namespace pollution
-Adheres to the C++ standard: Code adheres to C++ standard and avoids undefined behavior
-Appropriate Modularity: Code is grouped into appropriate modularity
-Uses modern C++ constructs: Code uses for loops and std::optional
-Uses standard library algorithms: no issue
-Pointer Handling: pointer uses raw pointers, but without new and delete. No use of either smart pointers std::unique_ptr and std::shared_ptr
-const correctness: no issue

Test:
-There is no testing for getNextRange, getPrevRange, and getContainingRange. Other than that, testing looks good and comprehensive!

Group-04/lib/CseString.hpp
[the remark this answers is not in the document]
PMitkiewicz: Overall, not very much for me to critique. It is well executed code, it compiles without warnings. Other than basic string, I think there are some other operations you might want with an audited string. For example, concat and replace are both strings I think you can audit, much like you did with substring. They should be relatively easy to add and will make the class even more useful. Other than that, great job!

132ikl:
@aydinorh 

> In line 139 instead of name 'a', you can use more descriptive name. Other than that, types and names are all good and understandable.

Thanks, resolved by 8d63777

> You can make these variables private and use getter functions to retrieve them

I don't think it's really necessary for these classes to have private members, since they're not intended to be used by downstream users anyways, as indicated by them being in the internal namespace.

> PrintAssertMessage and PrintLocation functions can be declared as const because they are not modifying the original class.

These functions aren't member functions of a class, so they can't be `const`.

> Also you should check for handling unexpected input types to catch misuses.

It's not really possible to add test cases for unexpected input types, since if the types aren't able to be compared the macro won't compile.

Thanks for your review!

Group-04/lib/CseAssert.hpp
```diff
@@ -0,0 +1,215 @@
...
+
+// These macros perform a best-effort recovery of the text of the original
+// arguments for display output on assertion failure.
+//
+// This will incorrectly split arguments in the same way that std's assert macro
+// does, but in this case only affects the display and not the logic. We could
+// potentially implement a parser to determine more accurately how the arguments
+// are split, but this is already an edge case.
+#define internal_cse_assert_args_head_(head, tail...) #head
+#define internal_cse_assert_args_tail_(head, tail...) #tail
```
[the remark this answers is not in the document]
132ikl: These macros specifically actually don't generate code directly, they just get the textual representation of arguments. I don't think the main macros don't need this either since they directly call C++ functions, so the semantics should be the same as calling a function normally. I do use a trick like this for the `dbg_assert` macros when they're disabled though, but I chose to use `(void)0` instead of `do { ... } while (0)` since I thought it was a bit cleaner.

Group-04/lib/CseAssert.hpp
```diff
@@ -0,0 +1,215 @@
...
+}
+
+void Assert(AssertArgs const &args, const char *args_text, const char *file,
+            int line, const char *function) {
+  if (args.condition) {
+    return;
+  }
+  std::string cond_string = std::string{args_text};
+
+  PrintAssertMessage(args.message);
```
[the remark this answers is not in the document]
132ikl: Great catch, thank you! Resolved by e9ae3a0

tests/Group-04/Assert.cpp
132ikl: Great ideas, added in 5ec19eb. Thanks!

Group-04/lib/CseAssert.hpp
```diff
@@ -0,0 +1,215 @@
...
+// are split, but this is already an edge case.
+#define internal_cse_assert_args_head_(head, tail...) #head
+#define internal_cse_assert_args_tail_(head, tail...) #tail
+
+// Internal macros which fill in the file, line number, etc. and call into the
+// actual C++ functions.
+#define internal_cse_assert_(condition_text, condition_msg...)             \
+  ::cse::_assert_internal::Assert(condition_msg, condition_text, __FILE__, \
+                                  __LINE__, __func__)
+#define internal_cse_assert_eq_(lhs_text, rhs_text, rest...)            \
```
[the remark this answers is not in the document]
132ikl: Addressed in unit tests comment

Group-04/lib/CseAssert.hpp
```diff
@@ -0,0 +1,215 @@
...
+  } else {
+    std::cout << "  These expressions evaluated to the same value:\n"
+              << "   left: " << lhs_string << "\n  right: " << rhs_string
+              << "\n";
+    PrintLocation(file, line, function);
+    std::cout << "hint: implement operator<< to see the underlying values\n";
+  }
+
+  Fail();
+}
```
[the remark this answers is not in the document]
132ikl: thanks :smile:

tests/Group-04/Assert.cpp
132ikl: This would definitely be a useful addition, but not in scope for a basic class I think. thanks though :heart:

Group-04/lib/CseAssert.hpp
```diff
@@ -0,0 +1,215 @@
...
+namespace cse::_assert_internal {
+#ifdef TEST_CSE_ASSERT
+class AssertTestException : std::exception {};
+void Fail() { throw AssertTestException(); }
+#else
+[[noreturn]] void Fail() { std::abort(); }
+#endif
+
+// This class allows us to accept both a condition and an optional message.
+//
```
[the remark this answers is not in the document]
132ikl: We didn't really decide on a specific style, and I figured the free-form explanation style comments would be more effective for this than Doxygen.

tests/Group-04/Assert.cpp
132ikl: I think that's kind of the point of the text in the first argument to test case. Did you have something specific in mind? While not super verbose, I do think the current explanations for each test case explain what's being tested.

Group-04/lib/CseString.hpp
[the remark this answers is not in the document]
132ikl: I moved the tests to catch2, thanks for the feedback

Group-04/bin/CseStringTest.hpp
132ikl: This ended up being removed since it's not relevant for catch2 tests

Group-04/lib/CseString.hpp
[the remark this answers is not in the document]
132ikl: Resolved by d8c38f58176c3d67fa8d6f49408e098e0d37ed17. Thanks for your review!

Group-04/lib/CseString.hpp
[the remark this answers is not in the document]
132ikl: > One suggestion I have for implementation is to add a little more details to error messages.

Krish addressed this in d8c38f58176c3d67fa8d6f49408e098e0d37ed17

Group-04/lib/BitVector.hpp
```diff
@@ -0,0 +1,854 @@
...
+}
+
+// Prepend another BitVector to the BitVector
+void BitVector::prepend(const BitVector& bv) {
+  BitVector tmp = *this;
+  *this = bv;
+  append(tmp);
+}
+
+BitVector& BitVector::operator+=(const BitVector& bv) {
```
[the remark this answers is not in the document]
132ikl: Resolved by d74c03a12ee74651ff1a249ad81adf9b4e07118b

tests/Group-04/BVOps.cpp
CircleShift: Added!

tests/Group-04/Assert.cpp
EvilJuiceBox, line 1: Well done with the testing here.

Group-04/lib/IndexSet.hpp
```diff
@@ -0,0 +1,573 @@
...
+ * consecutive numbers.
+ *
+ * The class automatically merges adjacent or overlapping ranges and splits
+ * ranges when removing elements. All ranges are stored in the half-open
+ * interval format [start, end) where end is exclusive.
+ */
+class IndexSet {
+ public:
+  // Iterator class definitions
+  class iterator {
```
mercere99, line 27: Nice addition of an iterator -- it makes this class much more usable.

Group-04/lib/IndexSet.hpp
```diff
@@ -0,0 +1,573 @@
...
+
+    bool operator!=(const iterator& other) const { return !(*this == other); }
+
+   private:
+    const std::vector<std::pair<std::size_t, std::size_t>>* ranges_;
+    std::size_t range_idx_;
+    std::size_t current_;
+  };
+
+  class const_iterator {
```
mercere99, line 87: Even better; a const iterator ensures that users of this class can easily respect const correctness.

Group-04/lib/IndexSet.hpp
mercere99: It does seem to be correct now; the comment should either be removed or clarified as to what the problem is.

Group-04/lib/IndexSet.hpp
```diff
@@ -0,0 +1,574 @@
...
+
+  /**
+   * @brief Adds an index to the set
+   * @param index The index to add
+   *
+   * If the index is adjacent to or within an existing range, the ranges will
+   * be automatically merged. For example, if [1,4) exists and we insert 4,
+   * the range becomes [1,5).
+   */
+  void insert(std::size_t index) {
```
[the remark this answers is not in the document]
mercere99: True, though since it's a pass-by-value and the full function is in the header file, it shouldn't make a difference (the compiler can identify that it doesn't change).  If either of those conditions weren't the case, it should definitely be marked `const`.

Group-04/lib/IndexSet.hpp
```diff
@@ -0,0 +1,573 @@
...
+    // First check if the index is already in a range to avoid duplicate work
+    auto range_idx = find_range_index(index);
+    if (range_idx < ranges_.size() && ranges_[range_idx].first <= index &&
+        ranges_[range_idx].second > index) {
+      return;  // Already in set
+    }
```
mercere99, line 195: I believe the code above could be simplified to:
```cpp
if (contains(index)) return; // Already in set.
```

Group-04/lib/IndexSet.hpp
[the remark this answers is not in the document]
mercere99: Also this function, and several others, can benefit from using helper functions, some of which already exist.  Right now it's longer and more unwieldy than it needs to be.

Group-04/lib/IndexSet.hpp
```diff
@@ -0,0 +1,573 @@
...
+
+    // Track old size before merging to determine if we need to increment
+    // total_size_
+    std::size_t old_total = 0;
+    for (const auto& range : ranges_) {
+      old_total += range.second - range.first;
+    }
```
mercere99, line 202: Don't you always increase total size by one since we know (from the previous code) that this index was not in there before.  It looks like you're doing a LOT of unnecessary work here (and below) to re-count total size.

Group-04/lib/IndexSet.hpp
```diff
@@ -0,0 +1,573 @@
...
+      total_size_ += range.second - range.first;
+    }
+
+    // If total size didn't increase after merging, we must have merged with an
+    // existing range In this case, we need to increment the size since we added
+    // a new index
+    if (total_size_ == old_total) {
+      total_size_++;
+    }
+  }
```
mercere99, line 223: The previous `for` loop just calculated total size manually.  How could it have gotten the size wrong?

Group-04/lib/IndexSet.hpp
```diff
@@ -0,0 +1,573 @@
...
+
+    // Merge overlapping ranges
+    merge_overlapping_ranges();
+
+    // Recalculate total size
+    total_size_ = 0;
+    for (const auto& range : ranges_) {
+      total_size_ += range.second - range.first;
+    }
+  }
```
mercere99, line 239: This seems so much more straightforward than you code to insert just one new index, which is odd.

Group-04/lib/IndexSet.hpp
```diff
@@ -0,0 +1,573 @@
...
+   * If the index is in the middle of a range, the range will be split.
+   * For example, if [1,5) exists and we remove 3, we get [1,3) and [4,5).
+   */
+  void remove(std::size_t index) {
+    // Find the range containing this index
+    auto range_idx = find_range_index(index);
+    if (range_idx >= ranges_.size() || ranges_[range_idx].first > index ||
+        ranges_[range_idx].second <= index) {
+      return;  // Index not in set
+    }
```
mercere99, line 254: Most times that you use `find_range_index` you then check to see if it's actually in the range or not.  My recommendation is that you have `find_range_index` do this check and return a special value if something is not in the code.

Basically, if you have:
```cpp
// Define the maximum size_t value to indicate that no range ID exists.
static constexpr size_t NO_RANGE = static_cast<std::size_t>(-1);
```
You can then make `find_range_index` reutrn NO_RANGE if the value was not in the set.  Your code above would then look like:
```cpp
auto range_idx = find_range_index(index);
if (range_idx == NO_RANGE) return; // index not in set
```
Many other places in this file could have a similar simplification.

Group-04/lib/IndexSet.hpp
```diff
@@ -0,0 +1,573 @@
...
+   * @brief Checks if an index is in the set
+   * @param index The index to check
+   * @return true if the index is in the set, false otherwise
+   */
+  bool contains(std::size_t index) const {
+    // Find potential containing range and verify index is within its bounds
+    auto range_idx = find_range_index(index);
+    return range_idx < ranges_.size() && ranges_[range_idx].first <= index &&
+           ranges_[range_idx].second > index;
+  }
```
mercere99, line 287: With the change to `find_range_index` I suggest above, this return statement would be just:
```cpp
return range_index != NO_RANGE;
```
In fact, you could simplify the whole function body to:
```cpp
return find_range_index(index) != NO_RANGE;
```

Group-04/lib/IndexSet.hpp
```diff
@@ -0,0 +1,573 @@
...
+
+  /**
+   * @brief Gets the next range after the given index
+   * @param index The reference index
+   * @return The next range if it exists, std::nullopt otherwise
+   *
+   * If the index is within a range, returns the next range after the
+   * containing range. If the index is between ranges, returns the next range.
+   */
+  std::optional<std::pair<std::size_t, std::size_t>> get_next_range(
```
mercere99, line 321: Using `std::pair<std::size_t, std::size_t>` for each range gets confusing.  I highly recommend creating a struct like:
```cpp
struct IndexRange {
  std::size_t start;
  std::size_t end;
};
```
This will make your code MUCH cleaner than needing to use `.first` and `.second` everywhere.  Seeing `IndexRange` as a type is also a LOT more informative than `std::pair<std::size_t, std::size_t>`.

Group-04/lib/RichText.hpp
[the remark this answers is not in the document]
mercere99: It does look like testing has been added now.

mercere99 (approved):
All-in-all some very nice code here; typically well thought out.  IndexSet has quite a lot that can be done to streamline its code, but overall functionality seems good.  I added many comments throughout.

Group-04/lib/BitVector.hpp
```diff
@@ -0,0 +1,885 @@
...
+constexpr size_t BITS_PER_EL = 8;
+
+// Author(s): Kyle Gunger
+// BitVector represents a dynamic array of bits, packed into bytes for space
+// efficiency. BitVector supports boolean operations on itself with other
+// BitVectors of varying sizes.
+class BitVector {
+ private:
+  // The m_underlying bytes of the vector
+  std::vector<std::byte> m_underlying = {};
```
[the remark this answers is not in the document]
CircleShift: Yeah that's a good idea.  I'll take a look into that.

tests/Group-04/BVSize.cpp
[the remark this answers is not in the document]
CircleShift: This was an artifact of me feeling like the BVOps file was getting too big, though they probably could be combined

Group-04/lib/BitVector.hpp
```diff
@@ -0,0 +1,885 @@
...
+  m_num_set += BIT_LOOKUP(m_underlying[idx]);
+
+  // Loop flip
+  for (++idx; (idx + 1) * BITS_PER_EL < start + count; ++idx) {
+    m_num_set -= BIT_LOOKUP(m_underlying[idx]);
+    m_underlying[idx] ^= flipper;
+    m_num_set += BIT_LOOKUP(m_underlying[idx]);
+  }
+
+  // Last byte
```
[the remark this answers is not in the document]
CircleShift: Yeah I can see that I was a bit too curt in some areas.  I'll go back and look over these

Group-04/lib/BitVector.hpp
```diff
@@ -0,0 +1,885 @@
...
+  (*this) = !(*this);
+  return *this;
+}
+
+//
+// BitVector functions
+//
+
+// Bit string constructor
+BitVector::BitVector(std::string bstr, char zero, char one) {
```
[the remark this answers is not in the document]
CircleShift: Thanks!  Honestly this function in particular didn't feel like my most clever, I feel most proud of the pattern_set utility.  I'm a bit curious, which part of this function did I write in a clever manner?

Group-04/lib/BitVector.hpp
```diff
@@ -0,0 +1,885 @@
...
+
+  // Bit string constructor
+  BitVector(std::string bstr, char zero = '0', char one = '1');
+
+  // Copy constructor
+  BitVector(const BitVector& bv) {
+    m_underlying = bv.m_underlying;
+    m_num_bits = bv.m_num_bits;
+    m_num_set = bv.m_num_set;
+  }
```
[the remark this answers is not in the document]
CircleShift: Yes this is true, I will change it.

Group-04/lib/BitVector.hpp
```diff
@@ -0,0 +1,885 @@
...
+    reference& operator=(bool value);
+    // Set the value of this bit
+    reference& operator=(const reference& value);
+    // Flip the value of this bit
+    reference& flip();
+
+    // Get the value of this bit
+    operator bool() const;
+    // Get the negated value of this bit
+    bool operator~() const;
```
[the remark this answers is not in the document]
CircleShift: Ah yeah, I probably just put it in when I was putting in the not operator for the bit vector and just thought "for parity" or something without thinking about it too deeply.  I'll remove this and add the other bit-wise setting operators as discussed below.

Group-04/lib/BitVector.hpp
```diff
@@ -0,0 +1,885 @@
...
+  };
+
+  // Default constructor
+  BitVector() = default;
+
+  // Constructor
+  BitVector(size_t bits) {
+    m_num_bits = bits;
+    m_underlying.resize(bits / BITS_PER_EL + (bits % BITS_PER_EL == 0 ? 0 : 1),
+                        ALL_ZERO);
```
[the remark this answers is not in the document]
CircleShift: Oh that's clever!  Yeah I'll look into factoring it out to a const function or something.

Group-04/lib/BitVector.hpp
```diff
@@ -0,0 +1,885 @@
...
+    // Constructor
+    reference(BitVector* bv, std::byte* b, uint8_t o)
+        : parent(bv), byte(b), bit_off(o) {}
+
+    // Set the value of this bit
+    reference& operator=(bool value);
+    // Set the value of this bit
+    reference& operator=(const reference& value);
+    // Flip the value of this bit
+    reference& flip();
```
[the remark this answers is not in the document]
CircleShift: Yeah this seems like a good idea, shouldn't take too much time and would make the reference class a little more useful

Group-04/lib/BitVector.hpp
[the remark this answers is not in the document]
132ikl: @CircleShift looks like you missed this one when going over reviews

Group-04/lib/CseAssert.hpp
```diff
@@ -0,0 +1,226 @@
...
+// dbg_assert_never doesn't uphold noreturn in release mode (since it is a
+// no-op), so this function acts like AssertNeverNoReturn but without the
+// [[noreturn]] marker
+void assert_never_return(
+    const char *file, int line, const char *function,
+    std::optional<std::string> message_opt = std::nullopt) {
+  assert_never_noreturn(file, line, function, message_opt);
+}
+
+}  // namespace cse::_assert_internal
```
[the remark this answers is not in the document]
132ikl: Thanks! We're actually thinking we will use Emscripten for our application, so I might try integrating something like this. Thanks for the suggestion

Group-04/lib/CseString.hpp
```diff
@@ -0,0 +1,194 @@
...
+   * @brief Overloaded operator[] (non-const) with debug checks.
+   * @param pos The index to access.
+   * @return Reference to the character at the specified position.
+   */
+  char &operator[](std::size_t pos) {
+    dbg_assert(pos < size(),
+               build_error_msg("cse::String operator[] index out of range", pos,
+                               size())
+                   .c_str());
+    return std::string::operator[](pos);
```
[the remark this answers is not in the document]
132ikl: Oh, this is actually from before I changed `AssertArgs` to have a `std::string` constructor. Thanks for pointing this out, resolved in 9d5f86a

Group-04/lib/RichText.hpp
```diff
@@ -0,0 +1,110 @@
+/**
+ * @file RichText.h
+ * @author Grant
+ *
```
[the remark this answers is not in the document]
132ikl: Resolved by f82319c

tests/Group-04/RichText.cpp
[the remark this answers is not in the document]
132ikl: > Are you testing for colours? Italics? 
I'm not sure exactly what you mean here, the formatting type is just a string, so downstream users can define formats however they would like (assuming it's compatible with their `OutputBackend`)

Group-04/lib/CseString.hpp
```diff
@@ -0,0 +1,194 @@
...
+  // const char*, size_t, etc.
+  using std::string::string;
+
+  /**
+   * @brief Explicit constructor from an existing std::string.
+   *
+   * This may help compilers that see std::__1::string as distinct
+   * from std::string in certain environments (e.g., macOS clang).
+   */
+  String(const std::string &other) : std::string(other) {}
```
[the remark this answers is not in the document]
132ikl: cc @KrishPatel198

tests/Group-04/RichText.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  cse::RichText::Format italic("italic");
+  text1.apply_format_to_range(italic, 4, 8);
+  REQUIRE(text1.formats_at(3).size() == 1);
+  REQUIRE(text1.formats_at(3).at(0) == bold);
+  REQUIRE(text1.formats_at(4).size() == 2);
+  REQUIRE(text1.formats_at(4).at(0) == bold);
+  REQUIRE(text1.formats_at(4).at(1) == italic);
+  REQUIRE(text1.formats_at(7).size() == 1);
+  REQUIRE(text1.formats_at(7).at(0) == italic);
```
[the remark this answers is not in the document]
132ikl: hmm... we don't actually have a way to remove a format. We'll need to add that for our application, so we can come back to this at that point.

PR #144, opened 2025-02-26
Group 5 Hamilton reference vector

greencalen17:


Group-05/src/ReferenceVector.h
```diff
@@ -135,7 +135,7 @@ class ReferenceVector {
       throw std::runtime_error("Index out of range.");
     }
 
-    assert(index < references_.size());
+    assert(index <= references_.size());
```
MaxKrawec, line 138: Shouldn't it be <? 

If [1,2,3]. Then the size is 3.
If you try to index at 3, it's an error.

greencalen17: If you were to insert the number 4 on a vector [1,2,3] at index 3 (4th element) it would just insert it to the back of the vector: [1,2,3,4] now 4 was inserted at index 4 which is valid. However, any further index would be a seg fault. So, I believe <= is correct and allows us to insert to the back of the vector. Here is a reference with some code you can edit and run:
https://www.geeksforgeeks.org/vector-insert-function-in-cpp-stl/

In most cases PushBack() will prolly be better but there may be strange cases where we want to insert to the back.

MaxKrawec: Makes sense. For some reason I thought you couldn't insert a value at the end (Like you said, most people use PushBack). Looks good then

greencalen17: Sweeet, thanks! Gonna merge now

PR #151, opened 2025-02-28
EventManager: Updated based on code reviews

graceKC04:
- Added RepeatEvent functionaity to update repeat if already repeated event is called
- Added comments to EventManager
- Added testcase for empty EventManager
- Added const to function parameters when applicable.


allienello (approved):
Everything looks good!

merryymary (approved):
Good job!

PR #163, opened 2025-03-13
Ritchie: Final Advanced Classes + Tests [StateGrid + AdvancedDataMap]

DominikLeisinger:


Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+  * @param name key name to be inserted
+  * @param val value to be associated with key within DataMap
+  */
+  template <typename T>
+  [[maybe_unused]] inline void insert(const std::string& name, const T& val) {
+    // https://stackoverflow.com/questions/3692954/add-custom-messages-in-assert custom messages in assert trick I have implemented within
+    assert(!contains(name) && "Key already exists within DataMap");
+    m_map[name] = Any(val);
+  }
+
```
allienello, line 318: Straightforward use of assert to enforce key uniqueness, nice approach

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+   * @tparam T any value
+   * @param name key name to be inserted
+   * @param val reference to be associated with key within DataMap
+   */
+  template <typename T>
+  [[maybe_unused]] inline void add_ref(const std::string& name, T& val) {
+    assert(!contains(name) && "Key already exists within DataMap");
+    m_map[name] = Any(val, true, &val);
+  }
+
```
allienello, line 330: Nice and clear support of inserting references; could possibly also use std::reference_wrapper as I mentioned in a previous comment

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+      return std::to_string(value);
+    } else if constexpr (has_to_string<T>) {
+      return value.ToString();
+    } else {
+      std::stringstream ss;
+      ss << value;
+      return ss.str();
+    }
+  }
+
```
allienello, line 393: Great use of checking if constexpr to handle different type conversions; maybe consider documenting the fallback behavior if no conversion condition applies

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+      if (pair.second.type() == typeid(T)) {
+        names.push_back(pair.first);
+      }
+    });
+    /// use for_each and specify a lambda within as a filter to grab desired keys
+    /// previous implementation was to use a for range loop instead of for_each
+    /// now captures names within the lambda and performs operation to push the string back within vector if type match
+    return names;
+  }
+};
```
allienello, line 513: Nice use of a lambda with std::for_each() to filter keys

allienello (commented):
The AdvDataMap class overall shows a solid grasp of advanced C++ techniques and good code design; it makes great use of compile-time type constraints, and it also has a smart type-erasure pattern with the inner Any class. That keeps the details hidden, adhering to encapsulation standards. The use of constexpr for type conversion was also cool to see,  though I think adding in a note on fallback behavior may be helpful. Switching from raw pointers to std::reference_wrapper for safer ref handling may also be beneficial but is also not necessary. Nice job group 2 :)

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
+struct StateDictionary {
+  std::map<char, cse::AuditedVector<std::string>> m_state_dictionary;
+
+  StateDictionary() {
+    m_state_dictionary = {{' ', {"EmptySpace", "Open"}},
+                          {'#', {"Wall", "Closed"}},
+                          {'X', {"Enemy", "Open"}},
+                          {'0', {"Exit", "Open"}},
+                          {'P', {"Player", "Closed"}}};
+  }
```
allienello, line 33: Nice initialization of the state dict with clear mappings

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
+    return m_state_dictionary.find(searchchar) != m_state_dictionary.end();
+  }
+  /**
+   * This function checks if a state can be traversed into
+   * @param searchchar state to be queried
+   * @return bool stating if the state is traversable
+   */
+  bool traversable(char searchchar) {
+    return m_state_dictionary.at(searchchar)[1] == "Open";
+  }
```
allienello, line 49: Good methods for checking state existence and traversability; use of map lookup keeps it efficient

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
+   * @param property The property to be changed/removed
+   * @param changeprop An optional variable, that can contain the new property or when not present, indicates removal
+   * The use of std::nullopt was suggested to me by ChatGPT
+   */
+  void change_property(char changestate, std::string& property, std::optional<std::string> changeprop = std::nullopt) {
+    auto it = m_state_dictionary.find(changestate);
+    if (it == m_state_dictionary.end()) {
+      return;
+    }
+
```
allienello, line 70: Nice use of std::optional for changeprop

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
-      {' ', {"EmptySpace", "Open"} },
-      {'#', {"Wall", "Closed"}},
-      {'X', {"Enemy", "Open"}},
-      {'0', {"Exit", "Open"}},
-      {'P', {"Player", "Closed"}}
-  };
- public:
+  cse::AuditedVector<std::string> m_grid;
+  ///StateDictionary object that holds all property info about StateGrid states
+  StateDictionary m_dictionary;
```
allienello, line 96: StateGrid does a great job of grouping grid data, state dict, and position into clear, separate parts

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
   }
 
-  [[maybe_unused]] explicit StateGrid(const std::string& diff);
+  /**
+ * @brief Contructor for StateGrid object
+ * @param diff string representing requested difficulty
+ */
+  explicit StateGrid(const std::string& diff) {
+    load_map(diff);
+  }
```
allienello, line 112: Great job using a default constructor to load a fallback map and a parameterized one for different difficulty levels

Group-02/src/cse/StateGrid.cpp
```diff
@@ -5,44 +5,46 @@
...
-StateGrid::StateGrid(const std::string& diff) {
-  load_map(diff);
+///NOTE: Used ChatGPT for syntax for this overload
+///Chat link: https://chatgpt.com/share/67d227cf-5f50-8001-86f4-494a0f71a7ea
+std::ostream& operator<<(std::ostream& os, const cse::AuditedVector<std::string>& grid) {
+  for (const auto& row : grid) {
+    os << row << "\n";
+  }
+  return os;
 }
```
allienello, line 23: Overloading operator << for the grid is a useful design choice for displaying its contents

Group-02/src/cse/StateGrid.cpp
```diff
@@ -5,44 +5,46 @@
...
-  if (validate_position({new_position.first, new_position.second})) {
-    m_grid[new_position.first][new_position.second] = 'P';
-    m_grid[agent.first][agent.second] = ' ';
+  if (validate_position({new_position.x_position, new_position.y_position})) {
+    m_grid[static_cast<size_t>(new_position.x_position)][static_cast<size_t>(new_position.y_position)] = 'P';
+    m_grid[static_cast<size_t>(agent.x_position)][static_cast<size_t>(agent.y_position)] = ' ';
+    m_position.set_object_position(new_position.x_position, new_position.y_position);
     return true;
   }
   return false;
```
allienello, line 50: Nice logic for updating the agent's position and grid state; use of static_cast for safe indexing is great

Group-02/src/cse/StateGrid.cpp
```diff
@@ -53,34 +55,65 @@ bool StateGrid::set_state(std::pair<int, int> new_position, std::pair<int, int>
...
 }
 
 /**
  * @brief Validates if a position can be occupied by agent
  * @param move std::pair containing row,col of position to be validated
  * @return T/F of position validity
  */
-bool StateGrid::validate_position(std::pair<int, int> move) {
+bool StateGrid::validate_position(std::pair<double, double> move) {
   assert(move.first < m_rows && move.second < m_cols && "This move is out of bounds");
```
allienello, line 114: Nice use of assert in validate_position and find_moves to check bounds

Group-02/src/cse/StateGrid.cpp
```diff
@@ -89,31 +122,59 @@ bool StateGrid::validate_position(std::pair<int, int> move) {
...
+ * @brief iterates through each cell and applies the passed function or lambda
+ * @param func function or lambda to be applied to each cell
+ */
+[[maybe_unused]] void StateGrid::modify_all_cells(const std::function<void(int, int, char&)>& func) {
+  for (int i = 0; i < m_rows; ++i) {
+    for (int j = 0; j < m_cols; ++j) {
+      func(i, j, m_grid[i][j]);
+    }
+  }
 }
```
allienello, line 179: Nice touch adding in a method that applies a lambda to each cell

allienello (commented):
The StateGrid class is organized very well and does a great job of separating the grid data, state dict, and agent's position. The default constructor loads a fallback map but also has a parameterized one for different difficult levels which is good code design. There is also solid logic for updated the agent's position, and safe indexing is done with static_cast. A previous review comment mentioned using operator << to display grid contents, which as mentioned would be helpful but is not necessary. Nice use of std::optional and lambda functions as well. Great job group 2 :)

graceKC04 (commented):
# Overall Comments:
- **Header File Type:**  
  Choose one file extension for header (`.hpp` is recommended for clarity the code is written in C++).



## StateGrid:

- **Implementation:**  
  The code is modular with the `StateDictionary` struct. Some functions could be marked as `const` (see line comments). Overall, it’s clean and easy to follow.

- **Advanced C++ Features:**  
  Consider using  `constexpr` in the `StateDictionary` constructor. Overall templating isn’t necessary here, and the current design works well.


## AdvancedDataMap:

- **Implementation:**  
  Modern C++ practices are followed, with `const` where appropriate. Memory management is handled carefully, and standard library functions are used effectively.

- **Advanced C++ Features:**  
  The `AdvancedDataMap` implementation uses modern C++ constructs such as templates in a clear and concise manner. The `Any` type is used to manage type-erased values, which allows for a highly flexible and type-safe interface.  
  **Lambdas** are used well in the `get_type_keys` method and serve a specific and clear purpose. 
  **`Constexpr`** is not widely needed here because most operations rely on runtime evaluation due to the dynamic nature of the data stored in the map.
  Overall, the advanced features like templates and type erasure are used effectively to create a very type flexible class , so additional comments are minimal as the code is already well developed.

tests/Group-02/ActionMapTest.cpp
```diff
@@ -3,8 +3,8 @@
  * @author Nitish Maindoliya
  */
 
-#include <catch2/catch.hpp>
 #include <cse/ActionMap.h>
+#include <catch2/catch.hpp>
 
 TEST_CASE("ActionMap: Intialization", "[ActionMap]") {
   SECTION("Default constructor") {
```
owenhaiar, line 10: For "ActionMap: insert, contains, erase, clear", add tests for checking the size of the map before and after modifying the map. Also check that the map is empty after removing all actions

tests/Group-02/ActionMapTest.cpp
```diff
@@ -60,11 +60,15 @@ TEST_CASE("ActionMap: insert, contains, erase, clear", "[ActionMap]") {
 }
 
 static int global_count = 0;
-void increment_global_count() { ++global_count; }
-void decrement_global_count() { --global_count; }
+void increment_global_count() {
```
owenhaiar, line 63: Good idea to test using a global var, I had the same idea

tests/Group-02/AdvDataMapTest.cpp
```diff
@@ -20,76 +20,314 @@ TEST_CASE("DataMap Default Constructor Test", "[DataMap]") {
...
+  cse::AdvDataMap data_map;
+  data_map["key"] = 80;
+  int value = data_map["key"].get<int>();
+  REQUIRE(value == 80);
+
+  data_map["key_two"];
+  auto val = data_map["key_two"].get<int>();
+  REQUIRE(val == 0);
+
+  SECTION("key not specified should not cause an error") {
```
owenhaiar, line 131: Consider also adding a test that a non-existent key returns the default value

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+      [[nodiscard]] virtual bool is_numeric() const = 0;
+      [[nodiscard]] virtual bool is_conv_to_string() const = 0;
+    };
+
+    template <typename T>
+    struct Val : Base {
+      /// store the value
+      T value;
+      /// flag to see if a reference is to be added to map
+      bool flag;
```
owenhaiar, line 63: This could use a more descriptive name

Group-02/include/cse/DataMap.hpp
```diff
@@ -60,7 +59,7 @@ class DataMap {
   * @param val value to be associated with key within DataMap
   */
   template <typename T>
-  [[maybe_unused]] void insert(const std::string& name, const T& val) {
+  inline void insert(const std::string& name, const T& val) {
     // https://stackoverflow.com/questions/3692954/add-custom-messages-in-assert custom messages in assert trick I have implemented within
     assert(!contains(name) && "Key already exists within DataMap");
```
miabodenhorn, line 64: Good assert, but also you could test for double insert attempts to make sure assert is triggered

Group-02/include/cse/DataMap.hpp
```diff
@@ -91,29 +90,38 @@ class DataMap {
    * @return T
    */
   template <typename T>
-  [[maybe_unused]] T get(const std::string& name) {
+  inline T get(const std::string& name) {
     if (!m_map.contains(name)) {
-      m_map[name] = std::any(T());
+      m_map[name] = std::any(T{});
     }
     assert(m_map[name].type() == typeid(T) && "Wrong type requested from what is contained within DataMap for value");
```
miabodenhorn, line 97: This assert is great but you should test what happens if you try to get a value with the wrong type, like storing an int but trying to get it as a std::string

tests/Group-02/AdvDataMapTest.cpp
```diff
@@ -20,76 +20,314 @@ TEST_CASE("DataMap Default Constructor Test", "[DataMap]") {
   }
 }
 
+TEST_CASE("DataMap Initializer List Constructor Test", "[DataMap]") {
+  cse::AdvDataMap data_map = {{"name", std::string("professor")}, {"age", 22}, {"dob", std::string("0120")}};
+
+  SECTION("DataMap is not empty") {
+    bool y = data_map.empty();
+    REQUIRE(y == 0);
```
owenhaiar, line 28: Checking a bool this way is kind of confusing, use REQUIRE_FALSE or y == False

tests/Group-02/AdvDataMapTest.cpp
```diff
@@ -20,76 +20,314 @@ TEST_CASE("DataMap Default Constructor Test", "[DataMap]") {
...
+  data_map.insert("key", 80.2);
+  auto val = data_map.get<double>("key");
+  REQUIRE(val == 80.2);
+}
+
+TEST_CASE("DataMap erase method Test") {
+  cse::AdvDataMap data_map;
+  data_map.insert("key", 80.2);
+  data_map.erase("key");
+  REQUIRE(data_map.empty() == 1);
```
owenhaiar, line 162: Same issue here, no reason to use an int when checking a bool

tests/Group-02/AdvDataMapTest.cpp
```diff
@@ -20,76 +20,314 @@ TEST_CASE("DataMap Default Constructor Test", "[DataMap]") {
...
+
+TEST_CASE("DataMap clear method Test") {
+  cse::AdvDataMap data_map;
+  data_map.insert("key", 80.2);
+  data_map.clear();
+  REQUIRE(data_map.empty() == 1);
+}
+
+TEST_CASE("DataMap size method Test") {
+  cse::AdvDataMap data_map;
```
owenhaiar, line 173: Also test sizes other than one and test the size again after a removal

tests/Group-02/AdvDataMapTest.cpp
```diff
@@ -20,76 +20,314 @@ TEST_CASE("DataMap Default Constructor Test", "[DataMap]") {
...
+
+TEST_CASE("DataMap to_string method Test") {
+  cse::AdvDataMap data_map;
+  data_map.insert("key", 80.2);
+  data_map.insert("name", std::string("professor"));
+
+  auto val = data_map.to_string<double>("key");
+
+  SECTION("Convert double to string") {
+    REQUIRE(typeid(data_map.to_string<double>("key")) == typeid(std::string));
```
owenhaiar, line 208: Since the return type of .to_string is guaranteed to be a string, this test doesn't do anything. Check that the return value is "80.2" instead.

tests/Group-02/AdvDataMapTest.cpp
```diff
@@ -20,76 +20,314 @@ TEST_CASE("DataMap Default Constructor Test", "[DataMap]") {
...
+  data_map.insert("key2", 10.10);
+
+  SECTION("Check keys with the type they should be") {
+    REQUIRE(data_map.is_type<int>("key1"));
+    REQUIRE(data_map.is_type<std::string>("key0"));
+    REQUIRE(data_map.is_type<double>("key2"));
+  }
+}
+
+TEST_CASE("Testing using an empty string as a key") {
```
owenhaiar, line 326: Good edge case. Maybe test that multiple values can be assigned to the empty string

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+
+  /**
+  * .at method for finding where key is
+  * @tparam T templated value returned
+  * @param name std::string
+  * @return
+  */
+  template <typename T>
+  [[maybe_unused]] inline T at(const std::string& name) const {
+    assert(contains(name) && "Key does not exist in DataMap");
```
owenhaiar, line 366: I would replace this assert with an exception, similar to how the std::map works. Use an assertion for the [] operator

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
+#include "StateGridPosition.h"
 
 namespace cse {
 
+/**
+ * StateDictionary
+ * Struct to hold each StateGrid state, along with any (and future) properties it has
+ */
+struct StateDictionary {
+  std::map<char, cse::AuditedVector<std::string>> m_state_dictionary;
```
owenhaiar, line 25: Since the entire class is in the cse namespace, no need to use cse::

Group-02/include/cse/AdvancedDataMap.hpp
merryymary, line 63: without the comment above (so when reading in the code), you can't tell what this flag is for. maybe rename to something more clear by itself.

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+     * Get stored value of type T from DataMap
+     * @tparam T type
+     * @return reference T of the value returned
+     */
+    template <typename T>
+    [[nodiscard]] const T& get() const {
+      using derive_t = Val<T>;
+      auto* derive_ptr = dynamic_cast<derive_t*>(value_ptr.get());
+      /// get raw pointer from value_ptr(unique ptr) and dynamically cast it (Base*) to a pointer of Val<T>
+      assert(derive_ptr != nullptr && "Wrong type requested from what is contained within Any");
```
merryymary, line 213: The assert with custom messages is a nice design choice for clarity.

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+    return m_map.empty();
+  }
+
+  /**
+    * Returns the count of the keys within the map
+    * @param name key to be found
+    * @return unsigned long # of keys
+    */
+  [[nodiscard]] [[maybe_unused]] inline size_t count(const std::string& name) const {
+    return m_map.count(name);
```
merryymary, line 479: From what I can understand, you can't have duplicate keys, so this count should really only be returning 0 or 1 for a specific key name. If this is the case, a simpler solution could be to just use contains or find to determine if a key exists.

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+  [[maybe_unused]] [[nodiscard]] inline bool is_ref(const std::string& name) const {
+    assert(contains(name) && "Key does not exist in DataMap");
+    return m_map.at(name).is_ref();
+  }
+
+  /**
+   * Emulate behavior of the [] operator as in std::unordered map
+   * Get value from key std::string
+   * If key does not exist, set key to defaulted any templated value constructor
+   * @tparam T templated value returned
```
merryymary, line 345: I noticed if a key doesn't exist that the default value gets assigned to it. I would think about if there would be any cases where a user would want to know if a key they were trying to get didn't exist, because with this a user may not know they were trying to access a previously non-existent pair.

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+   * @param name std::string
+   * @return T
+   */
+  template <typename T>
+  inline T get(const std::string& name) {
+    if (!m_map.contains(name)) {
+      m_map[name] = Any(T{});
+    }
+    assert(m_map[name].type() == typeid(T) && "Wrong type requested from what is contained within DataMap for value");
+    return m_map[name].get<T>();
```
merryymary, line 355: Overall this templating is really good for accessing values of many different types.

Group-02/src/cse/StateGrid.cpp
```diff
@@ -5,44 +5,46 @@
...
 
 /**
  * @brief Sets the specific position to occupied by agent
  * @param new_position pair containing new agent spot
  * @param agent pair containing current agent position
  */
-bool StateGrid::set_state(std::pair<int, int> new_position, std::pair<int, int> agent) {
-  assert(!m_grid.empty() && m_grid[agent.first][agent.second] == 'P');
+[[maybe_unused]] bool StateGrid::set_state(Point new_position) {
```
allienello, line 39: Nice use of safe indexing and assert here

merryymary:
Overall, StateGrid is a really solid class! The headers are very clear and all the same casing. I would potentially add underscores between each word in order to make the variable names more clear. Great job avoiding magic numbers. Member variables are correctly private and the functions are intuitive. Comments are correctly on function headers and describe sources as well :D The style is consistent within this class, but I did notice that other classes do the first_variable identifiers while this class has all one word. I would make sure all the classes have similar styles, not just within each class! This class seems very thought out and it is visually clear to read.

Group-02/include/cse/StateGrid.h
merryymary: I second this!

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
+
+  StateDictionary() {
+    m_state_dictionary = {{' ', {"EmptySpace", "Open"}},
+                          {'#', {"Wall", "Closed"}},
+                          {'X', {"Enemy", "Open"}},
+                          {'0', {"Exit", "Open"}},
+                          {'P', {"Player", "Closed"}}};
```
merryymary, line 32: Potentially, you could change from "open" and "closed" to just true or false, as its essentially check if its traversable or not. It would save maybe a few lines of code here and there, but I understand this way reading it is more clear. Just something to think about I guess!

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
+   */
+  bool find(char searchchar) {
+    return m_state_dictionary.find(searchchar) != m_state_dictionary.end();
+  }
+  /**
+   * This function checks if a state can be traversed into
+   * @param searchchar state to be queried
+   * @return bool stating if the state is traversable
+   */
+  bool traversable(char searchchar) {
```
merryymary, line 47: This is an example of the parameter name being searchchar but could be changed to search_char for better readability and consistency

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
```
merryymary, line 7: I believe these should be .hpp files not .h :D

Group-02/src/cse/StateGrid.cpp
merryymary, line 1: I don't see the specific line on here, but near line170 in modify_all_cells, it may be more flexible in the future to use iterators instead of just looping through indexes of the rows and columns of the map.

allienello (commented):
StateGrid is really well organized; it neatly separates the grid data, state dictionary, and the agent’s position. I like that it loads a fallback map and lets you choose different difficulty levels. The safe indexing with static_cast and clear update logic keep things reliable, and the use of lambdas and std::optional makes the code flexible. The tests back up the design by showing that grid states and property lookups work as expected. Great implementation that handles both the basics and edge cases smoothly.

AdvancedDataMap is a solid example of modern C++ design and implementation. It uses compile-time checks and a neat type-erasure design with its Any class, keeping the code modular and inside the cse namespace. I liked the templated constructor and if constexpr for type conversions; they make the code clear and safe. Smart pointers and if constexpr help keep the code clean, and the tests cover key operations like insertion, copying, and type conversion effectively; nice work.

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+    /**
+     * Get stored value of type T from DataMap
+     * @tparam T type
+     * @return reference T of the value returned
+     */
+    template <typename T>
+    [[nodiscard]] const T& get() const {
+      using derive_t = Val<T>;
+      auto* derive_ptr = dynamic_cast<derive_t*>(value_ptr.get());
+      /// get raw pointer from value_ptr(unique ptr) and dynamically cast it (Base*) to a pointer of Val<T>
```
miabodenhorn, line 212: Strong assert usage and clean dynamic casting

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+
+  /**
+   * Check if the value associated with the key specified is a reference
+   * @param name key name to be checked
+   * @return bool
+   */
+  [[maybe_unused]] [[nodiscard]] inline bool is_ref(const std::string& name) const {
+    assert(contains(name) && "Key does not exist in DataMap");
+    return m_map.at(name).is_ref();
+  }
```
miabodenhorn, line 339: You could add tests that try to access missing keys to verify assersions are triggered

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+    }
+    return sizeof(m_map.at(name));
+  }
+
+  /**
+    * Delete a key from the map
+    * @param name key to be deleted
+    */
+  [[maybe_unused]] inline void erase(const std::string& name) {
+    assert(contains(name) && "Key does not exist in DataMap");
```
miabodenhorn, line 446: Overall there is a nice and sonsistent use of assert!

Group-02/src/cse/StateGrid.cpp
```diff
@@ -5,44 +5,46 @@
...
 }
 
 /**
  * @brief Displays the grid to user
  */
 ///REVIEW COMMENT: an overloaded << to display StateGrid would be useful, but not needed
 ///unless we decide not to implement a better visual element to display our game.
-void StateGrid::display_grid() {
+[[maybe_unused]] void StateGrid::display_grid() {
   assert(!m_grid.empty() && "m_grid is empty and cannot display");
```
miabodenhorn, line 30: Nice safeguard before accessing or modifying the grid

Group-02/src/cse/StateGrid.cpp
```diff
@@ -89,31 +122,59 @@ bool StateGrid::validate_position(std::pair<int, int> move) {
...
-  } else {                                          ///<< If wrong map input, default to test map
-    m_grid = maps["test"];
+
+  ///Currently no check for incorrect filename other than thrown exception
+  std::string filename = diff + ".csv";
+  std::ifstream file(filename);
+  cse::AuditedVector<std::string> pregrid;
+
+  if (!file.is_open()) {
+    throw std::runtime_error("File could not be opened: " + filename);
```
miabodenhorn, line 153: I think you should possibly consider testing with an invalid file path to ensure this is triggered.

Group-02/src/cse/StateGrid.cpp
```diff
@@ -89,31 +122,59 @@ bool StateGrid::validate_position(std::pair<int, int> move) {
...
+ * @brief iterates through each cell and applies the passed function or lambda
+ * @param func function or lambda to be applied to each cell
+ */
+[[maybe_unused]] void StateGrid::modify_all_cells(const std::function<void(int, int, char&)>& func) {
+  for (int i = 0; i < m_rows; ++i) {
+    for (int j = 0; j < m_cols; ++j) {
+      func(i, j, m_grid[i][j]);
+    }
+  }
 }
```
miabodenhorn, line 178: Noticed here that m_grid[i][j] is accessed in modify_all_cells() without bounds checking, you might want ot assert that upfront

Group-02/src/cse/StateGrid.cpp
```diff
@@ -89,31 +122,59 @@ bool StateGrid::validate_position(std::pair<int, int> move) {
...
+  std::string line;
+  while (std::getline(file, line)) {
+    pregrid.push_back(line);
   }
+  file.close();
+
+  m_grid = pregrid;
   m_cols = static_cast<int>(m_grid[0].size());
   m_rows = static_cast<int>(m_grid.size());
+  m_position.set_object_position({1, 2});  ///< Until we implement more maps, hard coding original agent pos
```
miabodenhorn, line 165: no assert/validations for whether this point is inside the new map

Group-02/src/cse/StateGrid.cpp
```diff
@@ -5,44 +5,46 @@
...
 /**
  * @brief Sets the specific position to occupied by agent
  * @param new_position pair containing new agent spot
  * @param agent pair containing current agent position
  */
-bool StateGrid::set_state(std::pair<int, int> new_position, std::pair<int, int> agent) {
-  assert(!m_grid.empty() && m_grid[agent.first][agent.second] == 'P');
+[[maybe_unused]] bool StateGrid::set_state(Point new_position) {
+  auto agent = m_position.get_object_position();
+  assert(!m_grid.empty() &&
```
miabodenhorn, line 41: good use of assert to verify bounds

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+
+  template <typename T>
+  [[maybe_unused]] [[nodiscard]] inline std::vector<std::string> get_type_keys() const {
+    /// original method created by myself, refactored to utilizing a lambda within a for each
+    /// instead of a loop utilizing assistance by gpt4o
+    std::vector<std::string> names = {};
+    if (m_map.empty()) {
+      return names;
+    }
+    std::for_each(m_map.begin(), m_map.end(), [&names](const auto& pair) {
```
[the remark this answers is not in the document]
miabodenhorn: Took the words right out of my mouth! xD

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+        names.push_back(pair.first);
+      }
+    });
+    /// use for_each and specify a lambda within as a filter to grab desired keys
+    /// previous implementation was to use a for range loop instead of for_each
+    /// now captures names within the lambda and performs operation to push the string back within vector if type match
+    return names;
+  }
+};
+}  // namespace cse
```
miabodenhorn, line 514: All implementation is wrapped in namespace cse, nice job

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
+  bool find(char searchchar) {
+    return m_state_dictionary.find(searchchar) != m_state_dictionary.end();
+  }
+  /**
+   * This function checks if a state can be traversed into
+   * @param searchchar state to be queried
+   * @return bool stating if the state is traversable
+   */
+  bool traversable(char searchchar) {
+    return m_state_dictionary.at(searchchar)[1] == "Open";
```
miabodenhorn, line 48: Smart use of at() here for quick access and the "Open" check is simple and clear.

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
+  [[maybe_unused]] std::map<std::string, cse::AuditedVector<std::string>> find_properties();
 
-  std::vector<std::string> define_state(char state);
+  bool validate_position(std::pair<double, double> move);
 
-  bool validate_position(std::pair<int, int> move);
+  std::vector<Point> find_moves();
 
-  std::vector<std::pair<int,int>> find_moves(int row, int col);
+  [[maybe_unused]] void modify_all_cells(const std::function<void(int, int, char&)>& func);
```
miabodenhorn, line 137: Nice lambda use here!

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
 
+/**
+ * StateDictionary
+ * Struct to hold each StateGrid state, along with any (and future) properties it has
+ */
+struct StateDictionary {
+  std::map<char, cse::AuditedVector<std::string>> m_state_dictionary;
+
+  StateDictionary() {
+    m_state_dictionary = {{' ', {"EmptySpace", "Open"}},
```
miabodenhorn, line 28: i think this is a nice clean way to initialize all the state types in one place

tests/Group-02/StateGridTest.cpp
```diff
@@ -3,161 +3,141 @@
...
+#include <cse/StateGridPosition.h>
+#include <catch2/catch.hpp>
 #include <stdexcept>
 
-TEST_CASE("Testing StateGrid loads test map correctly")
-{
-  REQUIRE(1==1);
-  cse::StateGrid grid ("test");
+TEST_CASE("Testing StateGrid loads test map correctly") {
+  REQUIRE(1 == 1);
```
owenhaiar, line 12: Good edge case, if this fails we have a serious problem lol

Group-02/include/cse/StateGrid.h
owenhaiar, line 28: I would look into finding a way to use an enum class or something similar to hold this state information. A dict mapping strings to string tuples makes code that uses it harder to read

miabodenhorn (commented):
Implementation Quality:
Proper use of namespaces:
- There's no use of using namespace std; in headers or source files

C++ standard:
- No undefined behavior nice job!

Appropriate modularity:
- The design shows a high level of modularity:
- Helper structs like Val<T> encapsulate complex behavior.
- The StateGrid class breaks functionality across helper classes like StateDictionary and StateGridPosition.
- Functions are short and focused, and there’s minimal code duplication.

Modern C++ constructs:
- Good use of modern C++ features like std::optional, auto, range-based for loops, lambdas, std::unique_ptr and if  constexpr for type traits


Standard library algorithms:
- Not too much to say here, std::for_each and std::find are used effectively!

Pointer handling:
- Smart pointers like std::unique_ptr  are used properly in AdvDataMap::Any to manage dynamic memory

Const correctness:
- Member functions are appropriately marked const so nothing really to say here!


Testing & Error Checking
Unit testing:
- Tests are well organized 
- Constructors, copy/move semantics, insert, clear, erase, operator overloads, and type checks are all tested in AdvDataMapTest, nice work 
- StateGridTest covers map loading, movement logic, condition setting/removal, property lookup, and edge position checks.
- Some helper functions (like modify_all_cells) are tested both functionally and visually which is really nice

Edge cases:
- Great attention to edge case handling
- good file not found handling
- StateGrid tests grid boundaries and movement off the edge
- also more explicit testing of invalid file paths for StateGrid::load_map() would be useful

Assertions:
- Assertions are used effectively to enforce invariants
- overall i think these assertions are well placed 

Error Handling:
- Your error handling is solid but you might want to make some functions safer for users by providing alternatives that don’t crash the program when something goes wrong

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
+                          {'0', {"Exit", "Open"}},
+                          {'P', {"Player", "Closed"}}};
+  }
+  /**
+   * This function searches for a state in m_grid
+   * @param searchchar char that is being searched for
+   * @return bool stating if the state was found
+   */
+  bool find(char searchchar) {
+    return m_state_dictionary.find(searchchar) != m_state_dictionary.end();
```
owenhaiar, line 40: I'm kind of confused what this function is trying to accomplish. If m_state_dictionary is a constant mapping of states to properties, why are we checking if a state exists in that dictionary?

tests/Group-02/StateGridTest.cpp
```diff
@@ -3,161 +3,141 @@
...
+//}
+//REQUIRE(agentrow == "# P #");
+//}
+//SECTION("Testing Exit Placement")
+//{
+//REQUIRE(grid.get_state(4,1) == '0');
+//}
+//}
+
+TEST_CASE("Testing validate_position") {
```
owenhaiar, line 47: I think a lot of test cases are missing for stategrid, ie testing a move out of bounds, testing a move of more than one space, trying to move on a 1x1 grid, etc

Group-02/src/cse/StateGrid.cpp
DominikLeisinger: Thank you, it was suggested to me last code review and it makes it much simpler

Group-02/src/cse/StateGrid.cpp
DominikLeisinger: Thank you!

Group-02/include/cse/StateGrid.h
DominikLeisinger: Thank you!

Group-02/include/cse/StateGrid.h
DominikLeisinger: I had considered this, but I was waiting to see what other properties might come up in the future and I wanted to make sure in testing that the right property was getting indexed, thank you for the comment!

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
+#include "DataMap.hpp"
+#include "StateGridPosition.h"
 
 namespace cse {
 
+/**
+ * StateDictionary
+ * Struct to hold each StateGrid state, along with any (and future) properties it has
+ */
+struct StateDictionary {
```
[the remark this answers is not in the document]
DominikLeisinger: Thank you!

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+template <typename T>
+constexpr bool uses_to_string = requires(T & value) {
+  std::to_string(value);
+};
+
+template <typename T>
+constexpr bool has_to_string = requires(T & value) {
+  value.ToString();
+};
+
```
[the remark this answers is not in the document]
iamrb1: Thank you so much! I was trying to learn and play around with concepts like we learned in class.

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+ * Advanced version of DataMap class
+ * Features include: Custom memory management (custom any class)
+ * Type erased values paired to std::string keys
+ * Can add both values and references to DataMap
+ * Concept checks and type checks
+ */
+class [[maybe_unused]] AdvDataMap {
+ private:
+  /// Inspired by Dr.Ofria's in class Any implementation (refactored to use unique ptrs)
+  class Any {
```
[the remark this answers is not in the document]
iamrb1: Thank you! Appreciate the positive feedback.

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+    struct Base {
+      virtual ~Base() = default;
+      [[nodiscard]] virtual std::unique_ptr<Base> clone() const = 0;
+      [[nodiscard]] virtual const std::type_info& type() const = 0;
+      [[maybe_unused]] [[nodiscard]] virtual bool is_ref() const = 0;
+      [[maybe_unused]] virtual void set() = 0;
+      [[nodiscard]] virtual bool is_numeric() const = 0;
+      [[nodiscard]] virtual bool is_conv_to_string() const = 0;
+    };
+
```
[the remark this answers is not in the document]
iamrb1: Thank you for the positive feedback on the internals of my custom Any class!

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+    template <typename T>
+    struct Val : Base {
+      /// store the value
+      T value;
```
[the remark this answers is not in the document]
iamrb1: This is definitely much better to incorporate rather than a raw pointer and a flag. I just decided to do it this way as I set the value using the pointer according to the flag, which I think is faster but less safe.

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+      explicit Val(T value, bool flag = false, T* ref = nullptr) : value(value), flag(flag), ref_ptr(ref) {}
+
+      /**
+       * Clone function for copy assignment and constructor
+       * @return
+       */
+      [[nodiscard]] std::unique_ptr<Base> clone() const override {
+        return std::make_unique<Val<T>>(value, flag, ref_ptr);
+      }
+
```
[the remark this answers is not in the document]
iamrb1: Thank you! It was fun learning how to properly utilize advanced concepts in the creation of this class.

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+      using derive_t = Val<T>;
+      auto* derive_ptr = dynamic_cast<derive_t*>(value_ptr.get());
+      /// get raw pointer from value_ptr(unique ptr) and dynamically cast it (Base*) to a pointer of Val<T>
+      assert(derive_ptr != nullptr && "Wrong type requested from what is contained within Any");
+      if (derive_ptr->is_ref()) {
+        derive_ptr->set();
+        /// if the ptr is a reference as specified by the flag, i can set the value to the updated
+        /// value by dereferencing the ptr
+      }
+      return derive_ptr->value;
```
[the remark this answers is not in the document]
iamrb1: Thank you! This is heavily inspired by Dr. Ofria's method in class with a few minor tweaks (reference check + set).

Group-02/include/cse/AdvancedDataMap.hpp
iamrb1: Appreciate the feedback! As mentioned before, I am choosing to use my implementation for speed although std::reference_wrapper is a better option for safety.

Group-02/include/cse/AdvancedDataMap.hpp
iamrb1: Added change

Group-02/include/cse/AdvancedDataMap.hpp
iamrb1: Thank you!

Group-02/include/cse/AdvancedDataMap.hpp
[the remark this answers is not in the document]
iamrb1: Thank you both!

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+       */
+      [[nodiscard]] bool is_ref() const override {
+        return flag;
+      }
+
+      /**
+       * Dereference the ref_ptr and set it to value when the original is changed
+       */
+      [[maybe_unused]] void set() override {
+        if (ref_ptr) {
```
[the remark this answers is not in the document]
iamrb1: Accepted change

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+        value_ptr = nullptr;
+      }
+    }
+
+    /**
+     * Move constructor
+     * transfer ownership from from source to this
+     * @param in
+     */
+    Any(Any&& in) noexcept : value_ptr(std::move(in.value_ptr)) { in.value_ptr = nullptr; }
```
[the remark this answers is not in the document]
iamrb1: Thank you!

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
+/**
+* @file AdvancedDataMap.h
+* @author Rahul Baragur
+*/
+
+#pragma once
+
+#include <any>
```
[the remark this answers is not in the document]
iamrb1: accepted change

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+        /// if the ptr is a reference as specified by the flag, i can set the value to the updated
+        /// value by dereferencing the ptr
+      }
+      return derive_ptr->value;
+    }
+
+    /**
+     * Get the type of the stored any object
+     * @return const type_info reference
+     */
```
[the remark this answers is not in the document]
iamrb1: Thank you for the positive feedback on the concept checks!

Group-02/include/cse/AdvancedDataMap.hpp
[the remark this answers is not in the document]
iamrb1: Thank you for the positive feedback regarding the assert placement in my code!

Group-02/include/cse/AdvancedDataMap.hpp
iamrb1: accepted change, renamed to ref_flag

Group-02/include/cse/AdvancedDataMap.hpp
iamrb1: I first decided to use exceptions like you mentioned; however, in class we mentioned that it is usually a design choice when it comes to using asserts or exceptions. In this case (for my application) I prefer to use an assert but you are right: in a normal setting of the implementation of the DataMap, I would definitely use an exception instead.

Group-02/include/cse/AdvancedDataMap.hpp
iamrb1: Accepted change -> changed to ref_ptr

Group-02/include/cse/AdvancedDataMap.hpp
iamrb1: Thank you! The technique is very useful.

Group-02/include/cse/AdvancedDataMap.hpp
iamrb1: Accepted change -> Utilized find.

Group-02/include/cse/AdvancedDataMap.hpp
iamrb1: I tried to make this distinction between get and at. Get is supposed to emulate the [] operator since templates need to know at compile time what the type should be and since that cannot be done for the standard implementation of the [] operator, I decided to create the get method. At method is to be used if the user just wants to check like you mentioned.

Group-02/include/cse/AdvancedDataMap.hpp
iamrb1: Thank you for the positive feedback regarding templating in my code.

Group-02/include/cse/AdvancedDataMap.hpp
```diff
@@ -0,0 +1,514 @@
...
+  [[nodiscard]] [[maybe_unused]] inline size_t count(const std::string& name) const {
+    return m_map.count(name);
+  }
+
+  /**
+   * Operator overload for []
+   * @param name
+   * @return
+   */
+  Any& operator[](const std::string& name) {
```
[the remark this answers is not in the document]
iamrb1: Thank you! Unfortunately I could not perform the functionality that I desired for this operator overload so Get is a better choice to use. Nevertheless, I did my best with this overload.

Group-02/include/cse/AdvancedDataMap.hpp
iamrb1: Thank you!

Group-02/include/cse/AdvancedDataMap.hpp
iamrb1: Accepted change

Group-02/include/cse/AdvancedDataMap.hpp
iamrb1: Thank you!

tests/Group-02/AdvDataMapTest.cpp
```diff
@@ -20,76 +20,314 @@ TEST_CASE("DataMap Default Constructor Test", "[DataMap]") {
   }
 }
 
+TEST_CASE("DataMap Initializer List Constructor Test", "[DataMap]") {
+  cse::AdvDataMap data_map = {{"name", std::string("professor")}, {"age", 22}, {"dob", std::string("0120")}};
+
+  SECTION("DataMap is not empty") {
+    bool y = data_map.empty();
```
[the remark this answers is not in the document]
hoggbram: Yes that makes more sense, thank you

tests/Group-02/AdvDataMapTest.cpp
hoggbram: Yes that makes more sense, thank you

tests/Group-02/AdvDataMapTest.cpp
```diff
@@ -20,76 +20,314 @@ TEST_CASE("DataMap Default Constructor Test", "[DataMap]") {
...
+
+TEST_CASE("DataMap Edge Case", "[DataMap]") {
+  cse::AdvDataMap data_map;
+
+  SECTION("Access a key that is not in DataMap") {
+    REQUIRE_FALSE(data_map.contains("key"));
+  }
+}
+
+TEST_CASE("DataMap Copy Constructor Tests", "[DataMap]") {
```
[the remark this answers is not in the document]
hoggbram: Juts made this change, thank you

tests/Group-02/AdvDataMapTest.cpp
hoggbram: Just added this functionality, thanks

tests/Group-02/AdvDataMapTest.cpp
```diff
@@ -20,76 +20,314 @@ TEST_CASE("DataMap Default Constructor Test", "[DataMap]") {
...
+  data_map["key"] = 80;
+  int value = data_map["key"].get<int>();
+  REQUIRE(value == 80);
+
+  data_map["key_two"];
+  auto val = data_map["key_two"].get<int>();
+  REQUIRE(val == 0);
+
+  SECTION("key not specified should not cause an error") {
+    REQUIRE_NOTHROW(data_map["key_three"]);
```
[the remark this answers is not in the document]
hoggbram: Thanks, it would likely cause errors in the future if we didn't make sure this is consistently working

tests/Group-02/AdvDataMapTest.cpp
```diff
@@ -20,76 +20,314 @@ TEST_CASE("DataMap Default Constructor Test", "[DataMap]") {
...
+  data_map.insert("key", 80.2);
+  data_map.insert("key_two", std::string("value"));
+
+  SECTION("Success .at method") {
+    auto val = data_map.at<double>("key");
+    CHECK(val == 80.2);
+  }
+
+  SECTION(".at method wrong key specified") {
+    REQUIRE_FALSE(data_map.contains("key_three"));
```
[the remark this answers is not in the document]
hoggbram: I believe that all of our .at methods don't throw errors, and this is not compatible with the template structure we used for it. I agree that it could be something we improve on though, possibly by adding new functionality to support it

tests/Group-02/AdvDataMapTest.cpp
```diff
@@ -20,76 +20,314 @@ TEST_CASE("DataMap Default Constructor Test", "[DataMap]") {
...
+  }
+}
+
+TEST_CASE("DataMap get method Test") {
+  cse::AdvDataMap data_map;
+  data_map.insert("key", 80.2);
+  auto val = data_map.get<double>("key");
+  REQUIRE(val == 80.2);
```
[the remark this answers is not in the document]
hoggbram: We unfortunately cannot use throwing exceptions because we decided to go with asserts, so require_throws does not work on our end. The assert should just end the program if the wrong type is given

tests/Group-02/AdvDataMapTest.cpp
hoggbram: Changed it, thanks

tests/Group-02/AdvDataMapTest.cpp
hoggbram: Added, thank you for the suggestion

tests/Group-02/AdvDataMapTest.cpp
hoggbram: This is true, thanks for pointing that out

tests/Group-02/AdvDataMapTest.cpp
```diff
@@ -20,76 +20,314 @@ TEST_CASE("DataMap Default Constructor Test", "[DataMap]") {
...
+    REQUIRE(std::size(vector2) == 3);
+  }
+  SECTION("Check both at once") {
+    data_map.insert("key3", std::string("Test_key_3"));
+    data_map.insert("key4", std::string("Test_key_4"));
+    data_map.insert("key5", std::string("Test_key_5"));
+    data_map.insert("key6", std::string("Test_key_6"));
+    std::vector<std::string> vector21 = data_map.get_type_keys<std::string>();
+    std::vector<std::string> vector23 = data_map.get_type_keys<int>();
+    REQUIRE(std::size(vector21) == 5);
```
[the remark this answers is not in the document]
hoggbram: Yes that would make it much clearer

tests/Group-02/AdvDataMapTest.cpp
hoggbram: Good suggestion, especially for multiple types and such

Group-02/src/cse/StateGrid.cpp
```diff
@@ -53,34 +55,65 @@ bool StateGrid::set_state(std::pair<int, int> new_position, std::pair<int, int>
  * @param state queried state
  * @return vector of strings holding info about state
  */
-std::vector<std::string> StateGrid::define_state(char state) {
-  auto possible_state = m_dictionary.find(state);
-  assert(possible_state != m_dictionary.end() && "This state is not in the map!");
-  return m_dictionary.at(state);
+cse::AuditedVector<std::string> StateGrid::define_state(char state) {
```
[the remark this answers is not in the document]
DominikLeisinger: Youre correct, done!

Group-02/src/cse/StateGrid.cpp
```diff
@@ -53,34 +55,65 @@ bool StateGrid::set_state(std::pair<int, int> new_position, std::pair<int, int>
...
 /**
  * @brief Returns the state of a grid position
  * @param row row of position
  * @param col col of position
  * @return state of (row,col) position
  */
-char StateGrid::get_state(int row, int col) {
-  assert(row < m_rows && col < m_cols && "This is not inside the grid");
-  return m_grid[row][col];
+char StateGrid::get_state(Point statepos) {
```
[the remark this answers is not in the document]
DominikLeisinger: Correct, made the change, thank you

Group-02/src/cse/StateGrid.cpp
```diff
@@ -89,31 +122,59 @@ bool StateGrid::validate_position(std::pair<int, int> move) {
  * @param col col of agent position
  * @return vector of pairs of possible move directions
  */
-std::vector<std::pair<int,int>> StateGrid::find_moves(int row, int col) {
```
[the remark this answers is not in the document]
DominikLeisinger: Done, thank you

Group-02/src/cse/StateGrid.cpp
DominikLeisinger: There is a test in StateGridTest that makes sure the agent starts at [1,2]. I did not add any additional assurances for the correct position because there is only one grid file as of yet.

tests/Group-02/AdvStateGridTest.cpp
```diff
@@ -2,94 +2,193 @@
...
+  REQUIRE(grid.get_state(cse::Point(2, 2)) == 'X');
 
-    ///Possible new state (boulder that can be pushed)
-    REQUIRE(grid.get_state(2,6) == 'o')
-    ///Want to implement a liquid type (water,lava)
-    for(int i=0; i<6; i++)
-    {
-        REQUIRE(grid.get_state(6,i) == 'w');
-    }
+  // Row 0 is "#####"
```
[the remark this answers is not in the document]
hawkinsma: Thanks! We will add a comment showing what the test grid looks like

Group-02/src/cse/StateGrid.cpp
hawkinsma: Thanks, we will add a test to cover that

Group-02/src/cse/StateGrid.cpp
hawkinsma: Great idea, that would be a great use, we'll add more bounds checking.

tests/Group-02/StateGridTest.cpp
hawkinsma: Thanks for the specific suggestions, we'll add some more tests to cover these edge cases

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
 
 #pragma once
 
+#include <functional>
+#include <map>
```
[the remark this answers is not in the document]
DominikLeisinger: You are correct, we will include this, thank you

tests/Group-02/AdvStateGridTest.cpp
```diff
@@ -2,94 +2,193 @@
...
+TEST_CASE("Testing validate_position at boundaries") {
+  cse::StateGrid grid("test");
 
-TEST_CASE("Testing find_properties returns correct values")
-{
-    ///For now, and clarity of purpose, assuming find_properties returns
-    ///a map with the direction from agent as key, and a list of properties
-    ///of the state in that direction as the value
-    ///Using map info from above as context for testing
+  REQUIRE_FALSE(grid.validate_position({0, 0}));
```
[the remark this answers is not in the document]
hawkinsma: Good idea, we will add that test case, thank you!

tests/Group-02/AdvStateGridTest.cpp
```diff
@@ -2,94 +2,193 @@
...
-        ///Current agent location will be a member variable
-        auto props = grid.find_properties();
+  REQUIRE(grid.validate_position({1, 1}));
 
-        REQUIRE(props.at("DOWN") == {"Boulder", "Open", "Pushable"});
-        REQUIRE(props.at("LEFT")) == {"EmptySpace", "Open"});
-        REQUIRE(props.at("RIGHT")) == {"EmptySpace", "Open"});
-        REQUIRE(props.at("UP")) == {"Wall", "Closed"});
+  REQUIRE_FALSE(grid.validate_position({5, 4}));
+}
```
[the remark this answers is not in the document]
hawkinsma: Good ideas again, we will add more test cases

Group-02/include/cse/StateGrid.h
DominikLeisinger: Youre right! changes have been made.

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
+                          {'X', {"Enemy", "Open"}},
+                          {'0', {"Exit", "Open"}},
+                          {'P', {"Player", "Closed"}}};
+  }
+  /**
+   * This function searches for a state in m_grid
+   * @param searchchar char that is being searched for
+   * @return bool stating if the state was found
+   */
+  bool find(char searchchar) {
```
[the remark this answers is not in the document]
DominikLeisinger: Already done thanks to your comments, thank you

Group-02/include/cse/StateGrid.h
DominikLeisinger: Because there are other functions like define_state() and remove_conditions() that must ensure the requested state is in the dictionary before they return info about them/modify their properties

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
+  bool traversable(char searchchar) {
+    return m_state_dictionary.at(searchchar)[1] == "Open";
+  }
+  /**
+   * This function returns property information about a state in m_grid
+   * @param searchchar The state being queried
+   * @return AuditedVector of properties for queried state
+   */
+  cse::AuditedVector<std::string> get_info(char searchchar) {
+    return m_state_dictionary.at(searchchar);
```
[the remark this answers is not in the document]
DominikLeisinger: Done, thank you!

Group-02/include/cse/StateGrid.h
DominikLeisinger: I now see my inconsistency, I will make these changes for better readability, thank you!

Group-02/include/cse/StateGrid.h
DominikLeisinger: I agree since it is all C++, I refarctored my file, thank you!

Group-02/include/cse/StateGrid.h
```diff
@@ -7,56 +7,133 @@
...
 namespace cse {
 
+/**
+ * StateDictionary
+ * Struct to hold each StateGrid state, along with any (and future) properties it has
+ */
+struct StateDictionary {
+  std::map<char, cse::AuditedVector<std::string>> m_state_dictionary;
+
+  StateDictionary() {
```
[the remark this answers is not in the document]
DominikLeisinger: I do not believe a map or std::string are constexpr friendly(especially with custom Audited Vector class) when these variables are returned and continue outside of the function scope, and making it into an array with const chars would drastically change and complicate how to modify/lookup dictionary values. Thank you for your suggestion, if I am wrong I will certainly make this change.

Group-02/src/cse/StateGrid.cpp
```diff
@@ -89,31 +122,59 @@ bool StateGrid::validate_position(std::pair<int, int> move) {
...
   m_rows = static_cast<int>(m_grid.size());
+  m_position.set_object_position({1, 2});  ///< Until we implement more maps, hard coding original agent pos
+}
+
+/**
+ * @brief iterates through each cell and applies the passed function or lambda
+ * @param func function or lambda to be applied to each cell
+ */
+[[maybe_unused]] void StateGrid::modify_all_cells(const std::function<void(int, int, char&)>& func) {
+  for (int i = 0; i < m_rows; ++i) {
```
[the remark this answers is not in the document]
DominikLeisinger: I have changed the variable names to be more descriptive, however I like the idea of keeping the loops being limited by m_rows and m_cols as they are derived right from m_grid in load_map, but I will look into this and see if changing the function like this will be simpler and beneficial

Group-02/src/cse/StateGrid.cpp
DominikLeisinger: I like the idea of a possible change to modify_all_cells(), but if we were to change it I think a range-based loop using std::views would be the most likely change that would stay consistent to what wed like the function to look like

DominikLeisinger:
> Overall, StateGrid is a really solid class! The headers are very clear and all the same casing. I would potentially add underscores between each word in order to make the variable names more clear. Great job avoiding magic numbers. Member variables are correctly private and the functions are intuitive. Comments are correctly on function headers and describe sources as well :D The style is consistent within this class, but I did notice that other classes do the first_variable identifiers while this class has all one word. I would make sure all the classes have similar styles, not just within each class! This class seems very thought out and it is visually clear to read.

mercere99 (approved):
Nice work with your two class; `AdvDataMap` was especially well done.  Good use of advanced C++, including attributes.

Lots of comments throughout; in particular, Make sure to align header files to use the same extension, ideally `.hpp`.

PR #167, opened 2025-03-15
Templated Vertex, Edge, Graph, and GraphPosition

typritchett6:
Templated all the classes to have the type of vertex data and moved all related code to hpp files instead of cpp files for templating reasons. Also modified test cases to take templates into account.


gr005100 (approved):
besides conflicts well

LeoSSpecht (approved):

LeoSSpecht (approved):

PR #171, opened 2025-03-23
Liskov - Shift to single threaded and function based EventManager

graceKC04:
-Updated EventManager to be single threaded
-Updated EventManager to handle refactored events (events that contain templated functions)


merryymary (approved):
5 stars

allienello (approved):
Everything looks good!

PR #184, opened 2025-03-27
Bellman Advanced Classes

Devon-FoxElster:
The two Advanced Classes chosen are ComboManager and BruteForceOptimizer

- ComboManager now has a companion in the form of PermutationManager, which functions very similarly with the exception that it returns permutations rather than combinations
  - Both ComboManager and PermutationManager handle additional constrictions such as: 
    - Requiring a specific element in the given container
    - Allowing for the repeating of any elements, resulting in anywhere from 0 to k copies of that element
    - Also the combination of these two

- BruteForceOptimizer's changes were focused on providing a more optimized output compared to the original 'brute force' solution. We ran into a few issues for a while trying to give it speed increases, so any suggestions about how to further improve the current optimizations is very welcome
  - Current optimizations focus on stopping the recursion at depths where it is no longer possible for better values to be developed due to capacity limitations

- We also included updates from the 'Initial Classes' reviews into this PR alongside Advanced Class work.


tdkduong (commented):
Overall:
1. Code Optimization and Flexibility
- Consider applying concepts to constrain Container so that only types that meet specific requirements.
- Two constructors of ComboManager are too similar to each other
- The code could benefit from modern C++ range algorithms to reduce loops. For example, replacing manual loops in the propagation step with a range–based approach can improve clarity, such as
```c++
std::vector<ValueType> GetCurrentCombo() const {
  std::vector<ValueType> combo(indices_.size());
  std::ranges::transform(indices_, combo.begin(),
                         [this](std::size_t index) { return items_[index]; });
  
  if (isRequired_) {
    combo.push_back(requiredValue_);
  }
  return combo;
}
```
- If many of these calculations involve constant parameters, marking functions as constexpr would allow for compile–time computation when possible.

2. Code Design
- C++23 features like deducing `this` to simplify overloads and reduce redundant code between prefix and postfix operations, such as 
```c++
auto operator++(this auto& self) -> decltype(self)& {
  if (manager_.NextCombo()) {
    combination_ = manager_.GetCurrentCombo();
  } else {
    pastEnd_ = true;
  }
  pastBegin_ = false;
  return self;
}
```
- Throw Exceptions are better than Asserts in this code
- Variables like n_ and k_ are used for the total number of items and the combination size, but it might be unfamiliar for people who has not seen it before.

pinnama7:
Awesome job. Here is my code review, let me know your thoughts: https://docs.google.com/document/d/1kegMUyjkiVpSVpOAG59lvbF2XKtrsAcwF3KK_1ReZL8/edit?usp=sharing

tranbao01:
Implementation Quality:
No warnings during compilation: I compiled using Wall +Wextra. No warnings 

Proper uses of namespaces: All classes enclosed in cse; no namespace std 

Adheres to C++ standards: Code did compile cleanly

Appropriate Modularity: Both classes are modular
- BruteForceOptimizer uses SetItems, SetCapacity, and FindOptimalSolution.
- ComboManager has methods like NextCombo, PrevCombo, Reset

Modern C++ Constructs: Used Ranged based for loops, auto, templates and iterators for STL-like behavior in ComboManager

Standard Library Algorithms: Used mathematical reasoning (e.g., binomial coefficient C(n, k)) appropriately for Combo Manager.

PointerHandling: The code avoided using raw pointer

Const Correctness: The code did use const properly for functions like GetCurrentCombo, GetCurrentIndices and other parameters.

Testing and Error Checking:

Unit Testing:  The tests present all public member functions and are well organized
- ComboManager: Covered construction and setup, forward and backward iterations, edge conditions, advanced features.
- BruteForceOptimizer: Covered basic construction, setting capacity, knapsack scenarios, optimization and benchmarking.

Edge cases coverage: The tests did cover most edge cases
-ComboManager: Could include very large combinations (C(20,10)) to test performance (Idk probably non essential but still).
-BruteForceOptimizer: Solid edge cases coverage ( I can't think of anything else).

Assert Usage: assert() used properly throughout the hpp files, very solid.

Error Handling: ComboManager throws std::invalid_argument on bad indices; Tests catch and confirm exceptions

Group-10/Classes/BruteForceOptimizer.hpp
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  /**
+   * @brief Recursive helper that decides to include or exclude each item.
+   *
+   * @param index Index of the current item in the items vector.
+   * @param currentWeight The total weight of the selected items so far.
+   * @param currentValue The total value of the selected items so far.
+   */
+  void Search(size_t index, double currentWeight, double currentValue) {
```
nstanis21, line 101: Since you are calling the Search function so many times recursively, it may be more optimal to have currentValue and currentWeight as class variables that you add/subtract from for each recursion step instead of passing them through for each recursion step

Group-10/Classes/BruteForceOptimizer.hpp
```diff
@@ -0,0 +1,186 @@
...
+#include <utility>
+#include <vector>
+
+namespace cse {
+
+/**
+ * @brief Structure representing an item with a name, weight, and value.
+ */
+struct Item {
+  std::string name;
```
nstanis21, line 24: Could you make the name a Template type? It would allow for a user to use a smaller type than a string for the name if they want, and a user may want the optimization advantage of using an integer instead of a string if they are working with a large enough item set

Group-10/Classes/PermutationManager.hpp
```diff
@@ -0,0 +1,433 @@
...
+    return true;
+  }
+
+
+  /**
+   * @brief static helper function that is used to calculate permutation size
+   * @param x - the number we are performing the factorial operation on
+   * @return x!
+   */
+  static size_t Factorial_(size_t x) {
```
nstanis21, line 418: It may be worthwhile to store the results of factorial calculations using a map, along with the result of the largest calculated factorial, so if you later need a factorial result that was calculated as a step of a previous factorial, you can get it without having to recalculate it, and if you need to calculate a new largest factorial, you could skip some calculation steps by starting from where your previously largest factorial was instead from the beginning

nstanis21:
Testing & Error Checking:
BruteForceOpimizer had good test coverage, the only scenario that I found missing would be a test to make sure the class can't trigger an error from the recursion depth causing a stack overflow. One possibility could be using an assert to limit the size of the items_ vector to within a 'safe' size.

PermuationManager and ComboManager both also have good test coverage, but it might be worthwhile to implement tests using large 'random' sets, since they could find possible edge cases for you, and provide 'real use case' performance benchmarking. Also factorials can quickly grow very large, so it might be worthwhile adding an assert or check to prevent a value overflow error for the factorial function in PermutationManager

sidb70:
### Advanced class development

Templates - PermutationManager is implemented as class template to be flexible.

Value Semantics - Permutation Manager uses value semantics to copy containers passed by const reference. BruteForce optimizer uses value semantics throughout functions when storing and passing Items. 

Lambdas - used in setOptimizer function to sort by item weight


Overall, more advanced functionality could be added. There is one instance of templates and one instance of lambdas. Here are some examples that could be added:
- move semantics could be used by SetItems. 
- the optimizer only works on Item types. This could be templated to allow for various kinds of items with different scoring functions. 
- Factorial_ could be constexpr

### Implementation quality
Overall clean implementation. cse namespace is used throughout and the code is well-modularized with helper functions (NextKPermutation, Factorial, etc.). Good job using std algorithms for permutations and stable sort instead of writing from scratch.

improvements:
- Brute force optimizer: Search function has similar code for both the optimized and non-optimized cases (some lines within the conditional are duplicate). Consider making some helper function or refactoring this. 
- Combo: A lot of duplicate code in Prev() and Next(). Consider refactoring.

tdkduong:
Review for BruteForceOptimizer
## Code Design
### 1. Informative Identifiers
Good: BruteForceOptimizer, Item, SetItems, SetCapacity, and FindOptimalSolution are clear and self-explanatory.

Recommend: Rename Search to something more descriptive like ExploreCombinations or BacktrackSearch.

Recommend: Rename optimized_ to enableSearchPruning_ or similar for clarity.

### 2. Avoids Magic Numbers
Good: std::numeric_limits<double>::lowest() is properly used to represent an initial best score.

Recommend: Introduce constexpr constants instead of hardcoded values, especially for default limits or configurations.

### 3. Intuitive and Functional Interface
Recommend: Replace the mutable API (SetItems, SetCapacity, etc.) with a constructor that takes all required inputs. Example:

Good: Member variables are private, promoting good encapsulation.

### 4. Informative Comments
Recommend: Remove redundant comments that state the obvious (e.g., // BASE CASE...) and instead emphasize why decisions are made.

Recommend: Add more explanatory comments for optimization logic and pruning.

### 5. Consistent Style
Good: Clear and consistent usage of naming conventions, spacing, and indentation.

Recommend: Make setter names consistent—choose either PascalCase or camelCase for all (SetItems vs. setOptimizer).

### 6. Visual Appearance
Good: Overall clean layout with good use of whitespace.

Recommend: Reduce visual complexity in deeply nested logic by splitting into helper functions or early returns.

## Code Optimization & Flexibility
### 1. Search Algorithm Efficiency
Good: Backtracking is appropriate for small input sizes.

Recommend: Add memoization or branch-and-bound techniques to handle larger inputs more efficiently.

### 2. Nested Condition Simplification
Recommend: Refactor complex if-else branches in Search() into helper methods

### 3. Sorting Direction Bug
Recommend: Fix mismatch between comment and implementation in `setOptimizer()`:

Either change comparator to `a.weight > b.weight`, or

Update comment to say `"ascending" instead of "greatest to least"`.

### 4. Assert Usage
Recommend: Replace `assert(currentWeight >= 0)` with proper exception handling or precondition checks to make the code production-safe.

### 5. Scoring Function Flexibility
Recommend: Add support for a custom scoring function using `std::function<double(const Item&)>` to increase flexibility.

### 6. Avoid Redundant Computation
Recommend: Prevent `FindOptimalSolution()` from recomputing if inputs haven’t changed, or redesign to enforce a one-time solve per configuration.

tdkduong:
Review for BruteForceOptimizer
## Code Design
### 1. Informative Identifiers
Good: BruteForceOptimizer, Item, SetItems, SetCapacity, and FindOptimalSolution are clear and self-explanatory.

Recommend: Rename Search to something more descriptive like ExploreCombinations or BacktrackSearch.

Recommend: Rename optimized_ to enableSearchPruning_ or similar for clarity.

### 2. Avoids Magic Numbers
Good: std::numeric_limits<double>::lowest() is properly used to represent an initial best score.

Recommend: Introduce constexpr constants instead of hardcoded values, especially for default limits or configurations.

### 3. Intuitive and Functional Interface
Recommend: Replace the mutable API (SetItems, SetCapacity, etc.) with a constructor that takes all required inputs. Example:

Good: Member variables are private, promoting good encapsulation.

### 4. Informative Comments
Recommend: Remove redundant comments that state the obvious (e.g., // BASE CASE...) and instead emphasize why decisions are made.

Recommend: Add more explanatory comments for optimization logic and pruning.

### 5. Consistent Style
Good: Clear and consistent usage of naming conventions, spacing, and indentation.

Recommend: Make setter names consistent—choose either PascalCase or camelCase for all (SetItems vs. setOptimizer).

### 6. Visual Appearance
Good: Overall clean layout with good use of whitespace.

Recommend: Reduce visual complexity in deeply nested logic by splitting into helper functions or early returns.

## Code Optimization & Flexibility
### 1. Search Algorithm Efficiency
Good: Backtracking is appropriate for small input sizes.

Recommend: Add memoization or branch-and-bound techniques to handle larger inputs more efficiently.

### 2. Nested Condition Simplification
Recommend: Refactor complex if-else branches in Search() into helper methods

### 3. Sorting Direction Bug
Recommend: Fix mismatch between comment and implementation in `setOptimizer()`:

Either change comparator to a.weight > b.weight, or

Update comment to say "ascending" instead of "greatest to least".

4. Assert Usage
Recommend: Replace assert(currentWeight >= 0) with proper exception handling or precondition checks to make the code production-safe.

5. Scoring Function Flexibility
Recommend: Add support for a custom scoring function using std::function<double(const Item&)> to increase flexibility.

6. Avoid Redundant Computation
Recommend: Prevent FindOptimalSolution() from recomputing if inputs haven’t changed, or redesign to enforce a one-time solve per configuration.

Group-10/Classes/BruteForceOptimizer.hpp
Devon-FoxElster: Thank you for the comment, I did test this out on BruteForceOptimizer but ultimately I did not see any differences speed-wise for the algorithm. As we look for other methods of optimization we may revisit this though.

Devon-FoxElster:
Reply to @tdkduong 

The variables n_ and k_ were renamed to numItems_ and comboSize_ respectively. ManagerSetup_ has been eliminated due to constructor changes

I may need some clarification on your end about which default values you are referring to, as ComboManager/PermutationManager's bounds are heavily dependent on whatever user-defined container is given to us unfortunately. It is a similar case for BruteForceOptimizer

Ultimately the constructors got refactored into a single joint constructor. The boolean value for repeating elements necessitated that it was specified regardless of which constructor was used (size_t index getting implicitly converted to boolean instead of going to the required index). The use of std::optional was added to better specify when the user wants a required value and not.

Extraneous comments were removed where found

Standard C++ algorithms were converted to range algorithms where possible, and functions that could be constexpr or noexcept were labeled as such.

For assertions vs. throw, we decided on assertions for combination size not fitting the container as it is highly unlikely for the class to be called with a larger combo size than container size(or negative numbers, as they will overflow to high numbers) but we chose throws for handling improper required values.

I did look into deducing this for the operator overloads, but it doesn't seem to combine the postfix and prefix operators like your comment implied.

Devon-FoxElster:
Reply to @pinnama7 

Identifiers:
Weight and Value are names very commonly associated with the knapsack problem, which is what BruteForceOptimizer was designed to handle. The rationale for them is that each item has a value associated with it, but you can only 'carry' a set amount (hence weight).

I believe the rationale behind UpdateIndicesFrom_ was that the input was what we were updating from, like "Update Indices From <index>"
I think we will keep this as is, mainly due to it being readable in implementation as well as being a private function as opposed to public.

n_ and k_ have been renamed to more user-friendly names

From what I understand the choices between all .hpp vs a split approach can often be about taste, we ultimately chose all .hpp since it seemed to simplify templates + consistency

I'm not sure I agree with the lambda suggestions, primarily due to the length of the functions. The Search() function is the prime driver for the BruteForceOptimizer class, while BinomialCoefficient() for ComboManager is very long, and seems a bit unwieldy to place within a lambda.

Group-10/Classes/PermutationManager.hpp
Devon-FoxElster: I think this idea has merit, however currently the primary purpose of Factorial_ is upon initial construction where we calculate the total possible permutations. Due to the way PermutationManager works it will never need to be recalculated (as the initial container doesn't get changed unless a new PermutationManager instance is created)

Perhaps alongside the final application development we might look into a static variable within PermutationManager to propagate those results to all instances.

Devon-FoxElster:
Final review comments:
SetOptimizer's name has now been adjusted to fit the naming convention, and the problem comment within the function has been corrected

Search has been renamed to ExploreCombinations() and common checks have been pulled out to a helper function. Score Updates have been stored as lambdas within ExploreCombinations() to use when relevant

PermutationManager's Next() and Prev() functions have been refactored heavily to reduce repeated code, both now feed into a generic function to adjust the permutation.

Group-10/Classes/BruteForceOptimizer.hpp
```diff
@@ -0,0 +1,176 @@
...
+   */
+  void SetOptimizer(bool optimize) {
+    optimizeEnabled_ = optimize;
+    /// Sort greatest value first, with tiebreaker going to heavier items
+    if (optimizeEnabled_) {
+      std::stable_sort(items_.begin(), items_.end(),
+                       [](const Item& a, const Item& b) {
+                         return (a.value == b.value) ? a.weight > b.weight
+                                                     : a.value > b.value;
+                       });
```
mercere99, line 64: FYI, if you define the full set of comparisons for `Item` (as suggested above) then this would work without the lambda as:
```cpp
std::stable_sort(items_.begin(), items_.end(), std::greater<Item>());
```

Group-10/Classes/BruteForceOptimizer.hpp
```diff
@@ -0,0 +1,176 @@
...
+  double weight;
+  double value;
+
+  /**
+   * @brief == operator for two Items
+   * @return indication whether two Items are equal
+   */
+  friend bool operator==(const Item& a, const Item& b) {
+    return (a.name == b.name && a.weight == b.weight && a.value == b.value);
+  }
```
mercere99, line 35: Why is `operator==` marked `friend`?

If you want to super-simply define ALL of the comparison operators for `Item` you can use:
```cpp
auto operator<=>(const Item &, const Item &) const = default;
```
This would be instead of `operator==`, since it would cover that operator too.  It will compare the member variables in order, so in the current case it would compare `weight` first and move on to `value` only if `weight` was a tie.  Swap them above if you want them to be considered in the opposite order.

Group-10/Classes/PermutationManager.hpp
```diff
@@ -0,0 +1,424 @@
...
+   */
+  bool Next() {
+    /* Collect direction-specific checks (Previous vs Next) and call a
+       function with the shared code */
+    auto endCondition = [this]() {
+      if (!isRepeating_) {
+        if (currentIndex_ >= totalPermutations_ - 1) return true;
+      }
+      return false;
+    };
```
mercere99, line 120: Using this lambdas is a clever way to handle modularization.  Nicely done.

Group-10/Classes/PermutationManager.hpp
```diff
@@ -0,0 +1,424 @@
...
+    size_t res = 1;
+    for (size_t i = 2; i <= x; ++i) {
+      res *= i;
+    }
+    return res;
+  }
+
+  bool getPermutation_(auto endCondition, auto generatePermutation,
+                      auto updateRequiredIndex, auto failureCondition,
+                      bool isForward) {
```
mercere99, line 397: I like how you use auto here for the lambdas passed it; it will actually allow the lambdas to be inlined below, making for very efficient code.

If you wanted to get fancy, you could be using concepts / requires to ensure that the lambdas are the correct types, but I don't think that's critical.

mercere99 (approved):
Nice improvements to the `ComboManager` and `BruteForceOptimizer` classes.  Good updates based on reviewer comments.  I have a few inline comments, but nothing dramatic.

I look forward to your final app!

Group-10/Classes/BruteForceOptimizer.hpp
Devon-FoxElster: It was marked friend as a force of habit, I was used to doing it outside of the class itself. Thank you for the tip on operator<=>, we've redefined it as such for the Item struct.

Group-10/Classes/PermutationManager.hpp
Devon-FoxElster: Thank you!

PR #186, opened 2025-03-27
Group 4 Advanced Classes

132ikl:
The two advanced classes are `IndexSet` and `RichText`.


tests/Group-04/IndexSet.cpp
```diff
@@ -6,48 +6,61 @@
 
 // Test constructor and size calculation
 TEST_CASE("Constructor", "[IndexSetTest]") {
-  cse::IndexSet set1(0, 5);  // [0,5)
-  REQUIRE(set1.size() == 5);
+  // Index constructor
```
typritchett6, line 9: You have done a very good job testing all the constructors within both IndexSet and RichText. I especially like that you are making sure to test some of the less commonly tested constructors like Move and Range and also empty ranges.

tests/Group-04/IndexSet.cpp
```diff
@@ -6,48 +6,61 @@
...
 
   // Disjoint range
-  set.insert_range(12, 15);  // Should add [12,15)
+  set.insert_range(12, 15); // Should add [12,15)
   REQUIRE(set.size() == 13);
 
   // Invalid range (start >= end)
   set.insert_range(5, 3);
-  REQUIRE(set.size() == 13);  // Size should not change
+  REQUIRE(set.size() == 13); // Size should not change
```
typritchett6, line 55: I like that you are checking for invalid ranges. Should there be a certain trigger that occurs if you do try and insert an invalid range or should you potentially consider testing for REQUIRE_NOTHROW or some other error?

tests/Group-04/RichText.cpp
```diff
@@ -65,14 +71,14 @@ TEST_CASE("AppendFormattedText", "[RichTextTest]") {
...
 
-  cse::RichText::Format textSize("text size", 11);
-  text.apply_format_to_range(textSize, 0, 11);
+  cse::TextFormat textSize("text size", 11);
+  text.apply_format(textSize, 0, 11);
 
-  cse::RichText::Format link("link", "https://cse498.github.io/");
-  text.apply_format_to_range(link, 0, 11);
+  cse::TextFormat link("link", "https://cse498.github.io/");
+  text.apply_format(link, 0, 11);
```
typritchett6, line 81: What happens if apply_format is called with an out of bounds index range?

Group-04/lib/IndexSet.hpp
```diff
@@ -280,10 +395,14 @@ class IndexSet {
    * @return true if the index is in the set, false otherwise
```
typritchett6, line 395: (commenting here since I can't comment above on line 341). You could consider using cse_assert to validate valid input instead of just having a silent error you could test for it. Something like:
cse_assert(start < end && "IndexSet::insert_range called with invalid range");

tests/Group-04/IndexSet.cpp
```diff
@@ -192,76 +205,94 @@ TEST_CASE("SetOperations", "[IndexSetTest]") {
...
+    // Test equality
+    REQUIRE(cse::IndexSet{0, 3, 6} == cse::IndexSet{0, 3, 6});
+  }
+
+  SECTION("Test edge cases") {
+    cse::IndexSet empty_set;
+    REQUIRE(empty_set <= set1); // Empty set is subset of all sets
+    REQUIRE(set1 >= empty_set); // All sets are superset of empty set
+    REQUIRE((empty_set | set1).size() == set1.size()); // Union with empty set
+    REQUIRE((empty_set & set1).size() == 0); // Intersection with empty set
```
typritchett6, line 291: Very good job covering the edge cases present and testing for them to ensure correctness.

tests/Group-04/IndexSet.cpp
```diff
@@ -192,76 +205,94 @@ TEST_CASE("SetOperations", "[IndexSetTest]") {
...
+    auto diff_set = set1 - set2;
+    REQUIRE(diff_set == expected);
+
+    set1 -= set2;
+    REQUIRE(set1 == expected);
+  }
+
+  SECTION("Test symmetric difference") {
+    // Expected: [0,2) ∪ [3,5) ∪ [6,7)
+    // TODO: replace with equality check once adjacent insert bug is fixed
```
typritchett6, line 243: Has this bug been fixed and the code could be updated?

tests/Group-04/RichTextAdvanced.cpp
```diff
@@ -1,123 +1,102 @@
...
 
   REQUIRE(text2.to_string() == "hello, world!");
-  REQUIRE(text2.get_format_range(red) == std::optional{cse::IndexSet{0, 4}});
+  REQUIRE(text2.get_format_range(red) ==
+          std::optional{cse::IndexSet{std::pair{0, 5}}});
   REQUIRE(text2.get_format_range(blue) ==
-          std::optional{cse::IndexSet{7, text2.size() - 1}});
-  REQUIRE(text2.formats_at(5).size() == 0)
+          std::optional{cse::IndexSet{std::pair{7, text2.size()}}});
+  REQUIRE(text2.formats_at(5).size() == 0);
```
typritchett6, line 85: Could be useful to check an index before and after this to ensure you're meeting the boundary requirements.

Group-04/lib/IndexSet.hpp
```diff
@@ -292,16 +411,51 @@ class IndexSet {
...
+    return *result;
+  }
+
   /**
    * @brief Returns a vector containing all indices in the set
    * @return Vector of indices in ascending order
    */
   std::vector<std::size_t> get_all_indices() const {
     std::vector<std::size_t> result;
+
```
typritchett6, line 442: Would this function benefit from returning a const & to potentially help with how expensive large sets could be?

Group-04/lib/IndexSet.hpp
```diff
@@ -377,15 +531,63 @@ class IndexSet {
...
+   *
+   * @param shift_by The number of places to shift by
+   */
+  void shift_right(const std::size_t shift_by) {
+    if (auto max = max_index())
+      shift_right_within(shift_by, 0, *max + shift_by + 1);
+  }
+
+  /**
+   * @brief Adds another index set to the end of this one, starting at the
```
typritchett6, line 590: Could add an assert in append_at to make sure the offset isn't producing anything invalid.

Group-04/lib/RichText.hpp
```diff
@@ -17,109 +25,560 @@
...
+   */
+  [[nodiscard]] Underlying to_string() { return m_text; }
 
-  std::vector<Format> formats_at(size_t pos) const {
+  /**
+   * @brief Get the formats at a position
+   * @param pos The position to check for formats
+   * @return A vector of `TextFormat`s at position `pos`
+   */
+  [[nodiscard]] std::vector<TextFormat> formats_at(std::size_t pos) const {
```
typritchett6, line 271: You could consider returning by const & instead of value to avoid unnecessary copies.

tests/Group-04/RichTextAdvanced.cpp
```diff
@@ -1,123 +1,102 @@
...
-  cse::RichText<std::string> text1;
-  cse::RichText<std::wstring> text2;
+  // basic template (std::string)
+  cse::RichText text1;
 
-  // containers
-  cse::RichText<std::vector<int>> text3;
-  cse::RichText<std::vector<std::string>> text4;
+  // character type template
+  cse::BasicRichText<wchar_t> text2;
```
typritchett6, line 14: You do a very good job of testing for template flexibility throughout the tests.

tests/Group-04/RichTextAdvanced.cpp
```diff
@@ -1,123 +1,102 @@
...
-
-  // edge cases are tricky, we might just have to limit what is
-  // allowed, or reduce scope to exclude expanding formats (instead, just
-  // offseting them)
+TEST_CASE("RichText insertions and substitution", "[RichTextAdvanced]") {
+  cse::TextFormat red("color", "red");
+  cse::RichText text{"hello"};
+  text.apply_format(red, 0, text.size());
+
+  // ensure substitution retains formatting
```
typritchett6, line 93: Good job testing possible formatting issues like substitution and shifting.

typritchett6:
### **Advanced C++ Classes code review 2**
IndexSet and RichText are very well designed classes and have strong implementation and functionality. Overall a very strong submission and code is very readable!

Your unit tests are very well done. You test a large variety of constructors, edge cases, and other advanced behavior very well. You also do a very good job testing certain uncommon issues like format overlaps or character substitutions in RichText. I would consider adding some asserts in certain areas I commented on just because it could make the code more readable rather than just having something fail with no message. You could consider including some negative tests to try and catch any potential UB.

**Code Optimization and Flexibility**
You use a lot of good modern C++ concepts and design. IndexSet does a good job efficiently dealing with range merging and dealing with certain shiftings. RichText is also very well written. I would consider returning const & instead of returning vectors by value (mentioned inline code) to avoid unnecessary copying.

tests/Group-04/RichText.cpp
```diff
@@ -11,28 +11,34 @@ TEST_CASE("Constructor", "[RichTextTest]") {
...
+  cse::RichText text4(text3);
+  REQUIRE(text4.size() == 6);
+  REQUIRE(text4.to_string() == "string");
 }
 
 TEST_CASE("MultipleFormatting", "[RichTextTest]") {
   cse::RichText text1("hello world");
   REQUIRE(text1.size() == 11);
-  cse::RichText::Format bold("bold");
+  cse::TextFormat bold("bold");
```
LeoSSpecht, line 29: This is already easy to use and very good!
One suggestion that could make it even easier is to create some kind of default factories for common formattings like `bold`, `italic`, `underline` etc.

tests/Group-04/RichTextSerialize.cpp
```diff
@@ -0,0 +1,91 @@
...
+  text.apply_format(blue, 6, text.size());
+
+  cse::RichText::Serializer serializer{"HTML", header, footer};
+
+  auto color_start = [](cse::TextFormat const &format) {
+    cse_assert(std::holds_alternative<std::string>(format.metadata));
+    std::string color = std::get<std::string>(format.metadata);
+    return std::format("<span class={}>", color);
+  };
+  serializer.AddRule("color", color_start, "</span>");
```
LeoSSpecht, line 60: Very good flexibility allowing use of lambdas or strings on the Serializer.

tests/Group-04/RichTextSerialize.cpp
```diff
@@ -0,0 +1,91 @@
...
+  REQUIRE(output ==
+          (header +
+           "<span class=red>hello</span> <span class=blue>world</span>" +
+           footer));
+}
+
+TEST_CASE("Default HTML serializer", "[RichTextSerialize]") {
+  std::string header = "<!DOCTYPE html>\n<html>\n<head></head>\n<body>\n";
+  std::string footer = "</body>\n</html>\n";
+  auto serializer = cse::SerializerHTML();
```
LeoSSpecht, line 73: This is great design giving the user default parameters!
If this could be done for formatters also as I mentioned before it would be great!

Group-04/lib/RichText.hpp
```diff
@@ -17,109 +25,560 @@
...
+   */
+  TextFormat(FormatID name,
+             std::variant<std::string, int32_t, std::monostate> metadata)
+      : name(std::move(name)), metadata(std::move(metadata)) {}
+
+  /**
+   * Constructs a simple format without metadata
+   * @param name The name of the format
+   */
+  explicit(false) TextFormat(FormatID name)
```
LeoSSpecht, line 63: The `explicit` was new to me, but interesting keyword!

Group-04/lib/RichText.hpp
```diff
@@ -17,109 +25,560 @@
...
-  void apply_format_to_range(const Format& format, const size_t begin,
-                             const size_t end) {
+  /**
+   * @brief Append a string to RichText
+   * @param str The string to append
+   * existing character to new characters
+   */
+  template <typename Str>
+    requires std::derived_from<Str, std::basic_string<CharT>>
+  BasicRichText& append(const Str& str) {
```
LeoSSpecht, line 337: This is minor and more opinion based. But `append` and `push_back` are similar to me in functionality.
Maybe they could all have the same name and just specializations of each other, that way the user would only have to remember one way of adding to RichText and the internals would be handled in the class.

Group-04/lib/RichText.hpp
```diff
@@ -17,109 +25,560 @@
...
+   * @param begin Beginning of range to apply format to
+   * @param end End (exclusive) of range to apply format to
+   */
+  void apply_format(const TextFormat& format, const std::size_t begin,
+                    const std::size_t end) {
     dbg_assert(
         end >= begin,
         std::format("Format range ends after beginning, begin: {}, end: {}",
                     begin, end));
+    apply_format(format, std::pair{begin, end});
```
LeoSSpecht, line 385: This is more of a question than a suggestion.
Does using `std::pair` here mean the pair is converted to a IndexSet and call the overloaded function above?

Group-04/lib/RichText.hpp
```diff
@@ -17,109 +25,560 @@
...
+    bool operator<(const FormatSerializeTracker& rhs) const {
+      dbg_assert(iter != end,
+                 "FormatSerializeTracker compare failed: invalid iterator.");
+      return (*iter).first < (*rhs.iter).first ||
+             ((*iter).first == (*rhs.iter).first &&
+              (*iter).second < (*rhs.iter).second);
+    }
+  };
+
+  [[nodiscard]] SerializeResult serialize(const Serializer& serializer) const {
```
LeoSSpecht, line 464: This is a complicated function, and it is expected as it handles a lot of things. It does have some comments, but dividing it more functions or adding more comments could make it clearer for the reader how this works.

LeoSSpecht:
First of all, this is some very impressive code and functionality, congratulations!
Here is my overall view on both classes

## RichText
  This is a great class, and I find it particularly useful, specially the HTML converted.

### Code optimization and Flexibility
  I could not really find anything that could be very optimized. One possible, but not guaranteed, option would be to try to create a structure similar to an AbstractSyntaxTree of formattings, and maybe (just maybe), it could be optimized a little without having to move across characters and iterators. However a change like this would likely make other parts harder, like removing formatting and such.

  The code is very flexible, it has templates whenever necessary, allows the user to customize their formattings and add new types of output they might want (like Latex for example) without much work. It also provides some good defaults like `bold`, `italic`, `color`, etc. Some more examples of flexibility I like are: support of ranges to create text, different ways to clear the formatting (clearing the whole text, just a piece of text, a specific format), and how you can get information in different ways (get parts of the text with specific formatting).

  I could not find much use of constexpr, but I am not sure if it would fit in this case.

### Code Design
  This code is great, the functions are small, well commented and do simple tasks that add up. The classes are divided into smaller parts like Serializable, TextFormat and other utils.

  The only part I would change is the `serialize` function which was a bit big in my opinion, breaking it up could make it easier to read.

Group-04/lib/IndexSet.hpp
```diff
@@ -292,17 +464,52 @@ class IndexSet {
    */
   std::size_t size() const { return total_size_; }
 
+  /**
+   * @brief Returns the maximum index in the set
```
LeoSSpecht, line 468:
```suggestion
   * @brief Returns the minimun index in the set
```

Group-04/lib/IndexSet.hpp
```diff
@@ -292,17 +464,52 @@ class IndexSet {
...
+   */
+  std::optional<std::size_t> min_index() const {
```
LeoSSpecht, line 470: I am not sure this is possible. But as whenever an index is inserted the ranges are sorted for merging, maybe to optimize the code you could just get the first element here, assuming it is not empty?
Same but opposite for max.

Group-04/lib/IndexSet.hpp
```diff
@@ -396,7 +739,7 @@ class IndexSet {
 
     // Add all ranges from the offset set to this set
     for (const auto& range : offsetSet.ranges_) {
-      insert_range(range.first, range.second);
+      insert_range(range.start, range.end);
```
LeoSSpecht, line 742: I believe `insert_range` sorts the ranges. So maybe here we could add all the ranges, and only then do the merge related operations?
Not sure if it would break anything by doing so, but it could improve efficiency

Group-04/lib/RichText.hpp
```diff
@@ -17,109 +25,603 @@
...
```
gr005100, line 464: One suggestion is to clean up the richtext [[nodiscard]] SerializeResult serialize(const Serializer& serializer const function in richtext. It uses multiple while loops making the logic hard to follow. Id advise breaking up parts of the logic into helper functions.

tests/Group-04/RichText.cpp
```diff
@@ -11,28 +11,34 @@ TEST_CASE("Constructor", "[RichTextTest]") {
...
   std::string str2 = "bold";
-  cse::RichText::Format f2(str2);
+  cse::TextFormat f2(str2);
   REQUIRE(text1.formats_at(3).at(0) == f2);
 
-  cse::RichText::Format italic("italic");
-  text1.apply_format_to_range(italic, 4, 8);
+  cse::TextFormat italic("italic");
+  text1.apply_format(italic, cse::IndexSet{std::pair{4, 8}});
   REQUIRE(text1.formats_at(3).size() == 1);
```
gr005100, line 42: Everything looks good however id advise adding some comments,

tests/Group-04/RichTextSerialize.cpp
```diff
@@ -0,0 +1,91 @@
...
+TEST_CASE("Complex serialization", "[RichTextSerialize]") {
+  std::string header = R"(<style>
+.red {
+  color: #ff0000
+}
+.blue{
+  color: #0000ff
+}
+</style>
+<body>)";
```
gr005100, line 42: Everything looks good however id advise adding more comments.

tests/Group-04/RichTextSerialize.cpp
```diff
@@ -0,0 +1,91 @@
...
+  cse::TextFormat blue("color", "blue");
+
+  cse::RichText text{"hello world"};
+  text.apply_format(red, 0, 5);
+  text.apply_format(blue, 6, text.size());
+
+  cse::RichText::Serializer serializer{"HTML", header, footer};
+
+  auto color_start = [](cse::TextFormat const &format) {
+    cse_assert(std::holds_alternative<std::string>(format.metadata));
```
gr005100, line 56: I like this assert so that it makes sure theres a color tag.

gr005100 (commented):
Implementation Quality:

No warnings during compilation: No warnings on my end.

Proper uses of namespaces: All your code is within namespace cse { ... }. Your code avoids the generic use of  using namespace std as well it avoids ALL namespace pollution in header files that will spread to users of those files.

Adheres to the C++ standard: The code does avoids undefined behavior. 

Appropriate Modularity: Everything looks good. I dont see any unnecessary code duplication and you use multiple helper functions such as shift_within, merge_overlapping_ranges, and others these 2 for example help 
the merge for adjacent and overlapping ranges as well as handle shifting for both right and left use. Same for rich text. One suggestion is to clean up the richtext [[nodiscard]] SerializeResult serialize(const Serializer& serializer const function in richtext. It uses multiple while loops making the logic hard to follow. Id advise breaking up parts of the logic into helper functions. 

Uses modern C++ constructs: Everything looks good. Your code uses modern C++ techniques and range based for loops. I see the use of min and max index as well as pre next and prev range as well I see the use of multiple templated functions such as template <ShiftDirection direction> used to shift within left and right within the range. As well for rich text for example str, template <typename CharT = char, typename Underlying = std::basic_string<CharT>> to store rich text and insert different types into richtext. 

Uses standard library algorithms: Everything looks good. Your code avoids extraneous loops when equivalent functionality exists in the standard library such as sort and many others. 

Const correctness: Everything looks good. You use a const when needed. One suggestion id consider is add const for [[nodiscard]] std::vector<TextFormat> formats_at(std::size_t pos) const however it's not necessarly needed. 


Testing and Error Checking:

Unit testing: Everything looks good. Your unit tests are present for all public member functrions and are well organized. You test for multiple constructers and are well versed on top of that iterator functionality and multiple different operations. You also tests your advanced functions you have multiple tests and cases for constructions and more. One suggestion id advise is adding additonal comments in RichTextSerialize and RichText. 

Edge cases: Everything looks good. Your code tests for empty range, Invalid range (start >= end), testing difference, overlapping, erasing substrings, and much more which align covering most edge cases. 

Assertions: Your use of asserts are minimal however not needed unless to enforce invariants and ensure a valid state. I like how you use a assert cse_assert(std::holds_alternative<std::string>(format.metadata)); to create a color tag so that a color tag. 

Error handling: Everything looks good. You use a assert for creating a color tag to expect a string. You hanlde invalid operations and more. A suggestion would be to add some error handling for text size, out of bounds, potential range checks, could be helpful.

Overall everything looks good and follows the best practices covered through out the course and rubric. As mentioned through out this review id advise adding some additonal error handling, more helper functions for SerializeResult.

tests/Group-04/IndexSet.cpp
```diff
@@ -279,10 +310,20 @@ TEST_CASE("Test basic operations", "[IndexSetTest]") {
...
+    set.remove(5);
+    REQUIRE(set.size() == 0);
+    REQUIRE(!set.contains(5));
+  }
+
+  SECTION("Test additional insert") {
+    set.insert(7);
+    REQUIRE(!set.contains(4));
+    REQUIRE(set.contains(5));
+    CHECK(!set.contains(6)); // FIXME
```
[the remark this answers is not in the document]
132ikl: There was a bug, but it's fixed as of db6eb6338ec2bf037e45050150105828ddcf61a3

tests/Group-04/IndexSet.cpp
132ikl: Forgot to fix this as part of db6eb6338ec2bf037e45050150105828ddcf61a3, but fixed by cffda9f1152fc5b55706febe5dcf22164abf776b

Group-04/lib/IndexSet.hpp
132ikl: We have to allocate the vector within this function since the vector of indices has to be calculated from the current state of `IndexSet`, so we can't just reference pre-existing data unfortunately

Group-04/lib/RichText.hpp
132ikl: This isn't possible for the same reason as the `IndexSet`, the result vector has no place in memory so we can't return a reference to it. In reality this probably isn't a huge issue because of RVO

tests/Group-04/RichText.cpp
132ikl: I'm not sure how useful this would be because you still have to have to specify the types of format. Even if we made something like `cse::TextFormat::MakeBold()`, it would end up being a lot more verbose than just using the constructor:
```cpp
cse::TextFormat bold = cse::TextFormat::MakeBold();
```
and still the case even if you use `auto`:
```cpp
auto bold = cse::TextFormat::MakeBold();
```

Group-04/lib/RichText.hpp
132ikl: Yes

Group-04/lib/IndexSet.hpp
132ikl: thanks!

Group-04/lib/IndexSet.hpp
132ikl: Great point, changed in dc28794

Group-04/lib/RichText.hpp
132ikl: `append` is for appending an entire string, whereas `push_back` is for appending a single character (since strings are kind of like a container of characters). This is based off the API for `std::string`, which does the same thing ([`append`](https://en.cppreference.com/w/cpp/string/basic_string/append), [`push_back`](https://en.cppreference.com/w/cpp/string/basic_string/push_back))

KollinBartz (commented):
RichText:

Advanced C++ Features:
You used Lambdas, Lambdas, Value semantics, etc. You used a bunch and accurately throughout the code. I don't really see any places for improvement.

Implementation Quality:
The quality is insanely well made.
All review 1 criteria is met. I will comment in-line what I especially like.

IndexSet:

Advanced C++ Features:

You used Lambdas, Lambdas, Value semantics, etc. You used a bunch and accurately throughout the code. I don't really see any places for improvement.

Implementation Quality:
All review 1 criteria is met. I will comment in-line what I especially like.

For both classes I feel as though the code is the most thorough code I have reviewed in a long time (even professional settings if that tells you anything) I will leave code segments in-line where I can but the whole code is awesome. A+ from me.

Group-04/lib/RichText.hpp
```diff
@@ -17,109 +25,603 @@
...
-    auto iter = m_formatting.find(format);
+  /**
+   * @brief Get the range corresponding to a `TextFormat`
+   * @param format The format to get the range for
+   * @return The IndexSet corresponding to the format, if it exists
+   */
+  [[nodiscard]] std::optional<IndexSet> get_format_range(
+      const TextFormat& format) const {
+    const auto iter = m_formatting.find(format);
     if (iter == m_formatting.end()) return {};
```
KollinBartz, line 441: This is more of a formatting opinion, but I know y'all can use inline functions, statements and single line code. Y'all also do so beautifully, but you might want to consider avoiding them in order to keep consistency throughout the code for formatting. It doesn't really matter just an option.

Group-04/lib/RichText.hpp
KollinBartz: I think just adding more comments as I think the code itself is elegant.

Group-04/lib/RichText.hpp
```diff
@@ -17,109 +25,603 @@
...
+          reapply_rule_iter = tracker_iter;
+          ++reapply_rule_iter;
+        }
+        if (reapply_rule_iter == trackers.end()) break;
+        if ((*reapply_rule_iter->iter).first > current) break;
+        if ((*reapply_rule_iter->iter).second <= current) {
+          auto tracker_to_update = trackers.extract(reapply_rule_iter);
+          ++tracker_to_update.value().iter;
+          if (tracker_to_update.value().iter != tracker_to_update.value().end)
+            trackers.insert(std::move(tracker_to_update));
```
KollinBartz, line 562: Might want to consider adding more inline comments here just to break down even more what is happening as there is a decent sized portion with different objectives uncommented.

Group-04/lib/RichText.hpp
gringrant: This actually used to be a smart pointer to the object in memory, but we decided just returning it and letting the compiler optimize it was the best for what we were doing.

Group-04/lib/IndexSet.hpp
132ikl: @chrispaul-lab will be taking a look at implementing this

Group-04/lib/RichText.hpp
132ikl: @gringrant will be taking a look at this

tests/Group-04/RichText.cpp
132ikl: Made these tests a bit more readable/self-explanatory and added a comment in 5273a35

tests/Group-04/RichTextSerialize.cpp
132ikl: Fixed in d112ff8

Group-04/lib/RichText.hpp
132ikl: going to track this in [this related conversation](https://github.com/CSE498/CSE498-Spring2025/pull/186#discussion_r2020270348), marking this as resolved

Group-04/lib/RichText.hpp
132ikl: going to track this in [this related conversation](https://github.com/CSE498/CSE498-Spring2025/pull/186#discussion_r2020270348), marking this as resolved

Group-04/lib/IndexSet.hpp
132ikl: I don't think there's anything that the offset could produce which would be "invalid", and if it is then this should be covered by the `append_at` unit tests

tests/Group-04/RichTextAdvanced.cpp
132ikl: Fixed in 0abe2ff

Group-04/lib/IndexSet.hpp
132ikl: Fixed in ac89a95

tests/Group-04/IndexSet.cpp
132ikl: Fixed in ac89a95

tests/Group-04/RichText.cpp
132ikl: great point, fixed and added a test in d0e888b

Group-04/lib/RichText.hpp
```diff
@@ -17,109 +25,659 @@
...
+ * Known as filter map in other languages
+ */
+// we didn't end up needing this, but i'll leave it here anyway
+[[maybe_unused]] auto filter_transform(auto const& func) {
+  return std::views::transform(func) |
+         std::views::filter(
+             [](auto const& option) { return option.has_value(); }) |
+         std::views::transform([](auto const& option) { return *option; });
+}
+
```
132ikl, line 682: @mercere99 Here's a ~~nerd snipe~~ puzzle: is it possible for this function to `requires` a concept that ensures that `func` returns a `std::optional` without manually specifying the template parameters? I got it to work if you manually specify both the parameter type of the function _and_ the `value_type` of the `std::optional` as template parameters, but couldn't get the compiler to deduce them.

mercere99: I think it must be?  I can start by jumping back to the (now) old school, which I know better.  You can definitely make a type trait like:
```cpp
template <typename T>
struct is_std_optional : std::false_type {};

template <typename T>
struct is_std_optional<std::optional<T>> : std::true_type {};

// A simple helper
template <typename T>
inline constexpr bool is_std_optional_v = is_std_optional<T>::value;
```
(I can explain that more if you like, but that's a super-standard way to make a type trait -- just let the compiler pick the most constrained option that doesn't trigger a substitution failure.  Hrrrm...  maybe I should have taken some time this semester on traditional C++ metaprogramming; it looks like it might still be valuable.)

You can easily peel the return type off of a function `fun` with `decltype(fun())`.  So I think you can just use the type trait to require that to be optional?

If that doesn't work, I can play with it some more.  I'm sure it must be possible.

PR #187, opened 2025-03-27
Hamilton

PMitkiewicz:


MaxKrawec:
The advanced classes are DataGrid and ExpressionParser

hawkinsma:
I will be reviewing the testing and error checking as well as code optimization and flexibility for DataGrid. For the tests, your team did a great job, they were very extensive and thought of pretty much every corner case I could think of. The tests are well organized, cover all constructors and public member functions, as well as testing private member functions through  public ones. Assertions are present throughout a majority of the code, ensuring DataGrid is working correctly throughout runtime. The most fitting errors are thrown in functions where recovery is needed but an error has occurred, such as the at() method and throughout the statistical functions. The tests are neat and organized, with useful tags to trace the errors. The only issue I encountered when trying to run these tests were the constexpr functions for statistics. The error I received complained that a non-constexpr function std::vector::empty() was called in the constexpr functions. I am running C++ 20 and I know that constexpr is still limited for vectors, so that may be something to check on. I was able to run these tests with no errors once I removed the constexpr specifiers.

As for optimization and flexibility, again, great job here. Aside from double checking those constexpr functions, I only had a few suggestions in this area. First, in calculateStandardDeviation, the mean could be calculated once before the for loop, instead of each time. Also, if you had a large enough data grid, using an algorithm like QuickSelect for calculating the median could provide some performance increase, but on smaller grids, this probably wouldn't be very useful. Overall great job Group 5, especially in the test cases, very impressive.

Group-05/src/DataGrid.cpp
```diff
@@ -367,4 +468,519 @@ const DataGrid::row_t &DataGrid::at(std::size_t row_index_) const {
...
+
+/**
+ * Calculates the max value of the vector.
+ * @param double_values The given doubles vector.
+ * @return The max value of the vector
+ */
+constexpr double DataGrid::calculateMax(const std::vector<double>& double_values) const {
+  if (double_values.empty()) {
+    return std::numeric_limits<double>::quiet_NaN();
+  }
```
hawkinsma, line 829: These double_values.empty() calls are what caused the tests to initially fail, see review overview for details

Group-05/src/DataGrid.cpp
```diff
@@ -367,4 +468,519 @@ const DataGrid::row_t &DataGrid::at(std::size_t row_index_) const {
...
+ */
+constexpr double DataGrid::calculateStandardDeviation(const std::vector<double>& double_values) const {
+  if (double_values.empty()) {
+    return std::numeric_limits<double>::quiet_NaN();
+  }
+
+  double current_total = 0;
+  for (double value : double_values) {
+    current_total += std::pow(value - calculateMean(double_values), 2);
+  }
```
hawkinsma, line 800: The mean could be calculated prior to the loop once and reused

tests/Group-05/DataGridTests.cpp
```diff
@@ -1,161 +1,1459 @@
...
+  CHECK(not_equal_values_doubles_nan[1] == Datum(15.0));
+
+  cse::ReferenceVector<Datum> not_equal_values_strings_nan =
+      nan_grid.columnNotEqual(1, Datum("a"));
+  CHECK(not_equal_values_strings_nan.Size() == 1);
+  CHECK(not_equal_values_strings_nan[0] == Datum("e"));
+
+  cse::ReferenceVector<Datum> not_equal_values_nan =
+      nan_grid.columnNotEqual(0, Datum(std::numeric_limits<double>::quiet_NaN()));
+  CHECK(not_equal_values_nan.Size() == 3);
```
hawkinsma, line 1458: Great job on these tests, very in depth!

hawkinsma:
I will be also reviewing the testing and error checking as well as code optimization and flexibility for ExpressionParser. For testing, great job again, your team covered every public member function, as well as testing private functions through public members. The tests also directly test the constexpr functions as well as the MakeBinaryFun. I ran into one error when running the tests, it did not complete due to M_PI already being defined (probably in cmath from the header file). Once I removed the line with M_PI, it ran and passed all of the test cases though, so double check that simple issue (could just be on my end). There are a few well-placed asserts throughout the class, and I think more would be excessive, so good work. Errors are thrown with invalid arguments, which is an appropriate use case and logic errors are thrown automatically through the math operations. As for optimization and flexibility, I don't have a lot to add here, since this class is more straightforward but still very interesting and useful. The use  of std::function for each different mathematical function is a great use case and allows the class to be flexible in its use cases. Also, this class uses some templates and lambdas to make it even more general and reusable which is nice. Overall, great job!

tests/Group-05/ExpressionParserTests.cpp
```diff
@@ -3,40 +3,248 @@
 #include "../../Group-05/src/ExpressionParser.cpp"
 #include "../../Group-05/src/ExpressionParser.h"
 #include <map>
+#include <functional>
+
+constexpr double M_PI = 3.14159265358979323846;
```
hawkinsma, line 8: I got an error here for redefining M_PI when it is already in cmath, see overview comment for details.

MaxKrawec: Fixed (renamed the variable).
Thank you!

MaxKrawec: The bug is kinda weird. Another team member wrote the code without this PI constant, and everything worked for him because it was calling the cmath M_PI. However, when I tried to run it, my compiler couldn't find the M_PI variable in cmath. So I just added it as a constant with the same name (which in hindsight was a bad idea). It ran for me, but probably not for other people.

Group-05/src/DataGrid.cpp
```diff
@@ -1,37 +1,39 @@
...
-}
-
 /**
  * @brief Index a row of the DataGrid
  * @param row_index_ Index of the row
  * @return Indexed vector row from DataGrid
  */
 DataGrid::row_t &DataGrid::getRow(const std::size_t row_index_) {
-  return this->at(row_index_);
+  assert(row_index_ < grid_.size());
```
DominikLeisinger, line 30: Maybe consider an assert message, so if the program fails here you will be able to tell why right away

Group-05/src/DataGrid.cpp
```diff
@@ -295,13 +397,12 @@ void DataGrid::resize(std::size_t num_rows_, std::size_t num_columns_,
...
   os_ << "Grid Shape: " << rows << " x " << columns << "\n";
 
   for (std::size_t i = 0; i < rows; ++i) {
     for (std::size_t j = 0; j < columns; ++j) {
-      auto value = this->getValue(i, j);
-      if (value.IsString()) {
+      if (auto value = getValue(i, j); value.IsString()) {
         os_ << value.GetString() << " ";
       } else if (value.IsDouble()) {
         os_ << value.GetDouble() << " ";
```
DominikLeisinger, line 408: In the Datum &DataGrid::at() functions asserts, you access grid[0] without checking if the grid is empty first.

Group-05/src/DataGrid.h
```diff
@@ -28,24 +29,83 @@ class DataGrid {
   /// A 2D vector holding datums
   std::vector<row_t> grid_;
```
iamrb1, line 30: Informative Identifiers are clearly emphasized as variable names and functions are short and concise but highly descriptive and straightforward. For example, "grid" details the underlying data structure. Additionally, having an alias row_t promotes readability of code. Good job on emphasizing readability!

Group-05/src/DataGrid.h
```diff
@@ -28,24 +29,83 @@ class DataGrid {
...
+  };
+
+  /**
+   * Constructor using 2D Vector of Datums
+   * @param data
+   */
   explicit DataGrid(const std::vector<std::vector<Datum>> &data) {
-    assert(!data.empty());
-    assert(!data[0].empty());
+    assert(data.empty() || !data[0].empty());
```
iamrb1, line 73: I like that the prior asserts were concatenated into one assert. This is a good development practice especially when the asserts are related together. Great work!

Group-05/src/DataGrid.h
```diff
@@ -90,37 +150,108 @@ class DataGrid {
...
   [[nodiscard]] const Datum &getValue(std::size_t row_index_,
-                        std::size_t column_index_) const;
+                                      std::size_t column_index_) const;
+  [[nodiscard]] std::vector<row_t> getDataGrid() const { return grid_; };
+
   [[nodiscard]] std::tuple<std::size_t, std::size_t> shape() const;
-  cse::ReferenceVector<Datum> getColumn(std::size_t column_index_);
+
   void insertDefaultRow(
       std::size_t row_index_ = std::numeric_limits<std::size_t>::max(),
```
iamrb1, line 186: `std::numeric_limits<std::size_t>::max()` is repeatedly used and since this is known at compile time, you could possibly create a constant variable at the top of the file to eliminate redundancy. See if this change if possible!

Group-05/src/DataGrid.h
```diff
@@ -28,24 +29,83 @@ class DataGrid {
...
+  };
+
+  [[nodiscard]] std::vector<double>
+  getDoubleValues(const ReferenceVector<const Datum> &reference_vector) const;
+
+  [[nodiscard]] ReferenceVector<Datum> determineColumnComparisons(
+                                                    size_t column_index,
+                                                    const Datum &value,
+                                                    operations operation);
+
```
iamrb1, line 49: Encapsulation is highly prioritized with only necessary functionality given. Any member variables (such as grid_) and behind the scene methods (such as calculateMean) are marked under private so the user cannot access them directly. Great work emphasizing this!

Group-05/src/DataGrid.cpp
hoggbram, line 1: Advanced C++ Features:
I think you did a fantastic job implementing some of the advanced features we discussed in class. To start, many functions in the header are marked as nodiscard to ensure that they are actually being used correctly within the class. The use of explicit is also very smart to add. In merge, you use a pointer to the DataGrid object to copy it and update some values, which I think is a good use of pointers, especially because it can manage its own memory and doesn’t need a smart pointer. SortColumn uses many lambda functions very well, returning something as a parameter of another sort function which is a perfect use case for a lambda function.  Same case for the sort function you created. I think all the standard library functions you used was smart as well, to avoid unnecessary extra functions. Some examples you used are std::isnan, operations::lessthan, std::numeric_limits, etc. Overall very well done.

hoggbram, line 1: Implementation Quality:
Everything is contained within the cse namespace. I do not see using namespace std which is a plus as well. I like how you used helper classes with datum and referencevector. I also noticed a few helper functions such as clear and resize/insert. These appear to be overloaded which is a good way to handle multiple types as well. There are a lot of standard library functions used to avoid range based for loops, such as ranges::all_of within is_rectangle. Good use of nodiscard and explicit as discussed in class in functions such as is_rectangle and the constructors for DataGrid and referencevector. IsRectangle returns a check for emptiness twice, you could probably remove the first check. Const correctness looks good as the functions marked constexpr are also const. I like how many of the function parameters are marked as const, as it likely makes it easier to use them and not worry about unknown changes to variables. I believe this line could return a const instead of a copy though: [[nodiscard]] std::vector<row_t> getDataGrid() const { return grid_; };. This code is also commented and organized very well.

Group-05/src/ExpressionParser.cpp
hoggbram, line 1: Advanced C++ Features:
This class uses lambda functions very effectively, mainly in the functions that do mathematical equations. It is a good use of them as it only executes them when they’re necessary. Especially in makeFunc, it uses recursion to find the next key and many of the functions called here use lambdas. I would suggest to check for division by zero in MakeDivFun to avoid any errors or undefined behavior later on. Another course aspect used here very well is recursion. We haven’t talked a lot about recursion in class to my knowledge, but this class uses it in a very efficient way. Especially in MakeFunc, it uses recursion to create different functions based on what it returns from itself. Another thing that was done well here was the use of constexpr. It is used for the mathematical functions such as Add and Subtract. This is very useful as it can be done at compile time rather than runtime. The only thing I would point out here is that I believe exceptions  cannot be throws at compile time, so the constexprDivided function wouldn’t have much use being constexpr. For the functions addition, subtraction, division, multiplication, and exponent that are marked as constexpr, they don’t have a const reference as their parameters, so they cannot be done at compile time. Last concept I saw was the use of templates. The use of templates in Lessthan and Equalto was a good addition as the comparisons can be done at compile time.
To wrap it up, I believe there should be some asserts in here to do error checking in addition to the tests.

hoggbram, line 1: Implementation Quality:
This class is very well structured overall. All of the code is correctly within the namespace cse tag. I believe you could use an unordered_map for faster runtime if the order isn’t relevant in your application. There is proper usage of templates and constexpr throughout. There is also good use of recursion, especially in the MakeFunc function. The const correctness seems to be right from my understanding. Again, there probably should be more asserts throughout the class. There are very few loops, and the ones that do exist cannot be replaced with std::fill or other standard library functions which is good. Overall, this class is also very well commented and I felt like I understood the use of every function.

Group-05/src/DataGrid.h
```diff
@@ -90,37 +150,108 @@ class DataGrid {
...
       std::size_t row_index_ = std::numeric_limits<std::size_t>::max(),
       double default_value_ = 0);
   void insertDefaultRow(
       std::size_t row_index_ = std::numeric_limits<std::size_t>::max(),
       std::string default_value_ = "");
+
   void insertDefaultColumn(
       std::size_t column_index_ = std::numeric_limits<std::size_t>::max(),
       double default_value_ = 0);
   void insertDefaultColumn(
```
iamrb1, line 195: I love that the functions are overloaded to correctly reflect inserting double vs string and default arguments are provided proving a robust and functional interface for the user.

Group-05/src/DataGrid.cpp
```diff
@@ -156,23 +181,23 @@ void DataGrid::insertDefaultRow(std::size_t row_index_, double default_value_) {
...
-  std::size_t num_columns = std::get<1>(this->shape());
+  const std::size_t num_columns = std::get<1>(shape());
 
   // Source ChatGPT
   // Inserts row at row_index_
-  grid_.insert(grid_.begin() + static_cast<std::ptrdiff_t>(row_index_),
-               std::vector<Datum>(num_columns, Datum(std::move(default_value_))));
+  grid_.insert(
+      grid_.begin() + static_cast<std::ptrdiff_t>(row_index_),
+      std::vector<Datum>(num_columns, Datum(std::move(default_value_))));
```
iamrb1, line 200: I think this is a great place to add a detailed comment explaining the functionality for the readers' understanding. For example, it would be helpful to know the detailing behind casting row_index_ to a std::ptrdiff_t and why that is necessary. I think in general anything you generate from GPT would be useful to explain in a comment as some of the code might not be visually making sense to the average reader.

Group-05/src/DataGrid.cpp
DominikLeisinger, line 1: **Implementation Review for DataGrid:**
This .cpp and .h file clearly show your extensive knowledge of std:: algorithms and modern C++ constructs. I saw some standard library algorithms that I have not seen before, so it was great to see and learn something new. (e.g std::numeric_limits, std::ptrdiff_t) The use of constexpr in calculating mean, median, stdev, etc. show that you understood what was taught to us in class and were used intelligently. The functions were also organized very well, as all related functions were right next to the others. All functions that could have been made const were correctly done as such. One thing I did not understand were the couple functions like getRow, getColumn had one const version and one without const. I assumed this was an advanced version of the function as an indexing function should be made const? Overall, this file was extremely detailed and documented well, and showed the amount of planning/time that went into it. Great Job.

Group-05/src/DataGrid.cpp
```diff
@@ -130,18 +156,17 @@ cse::ReferenceVector<Datum> DataGrid::getColumn(const std::size_t column_index_)
  * @param default_value_ of row values
  */
 void DataGrid::insertDefaultRow(std::size_t row_index_, double default_value_) {
```
iamrb1, line 158: All methods are in camelCase, providing a standard code style for DataGrid. As mentioned in the general review, ExpressionParser contains PascalCase for some methods in which maybe you all could coordinate and have one set code style for the group.

Group-05/src/ExpressionParser.cpp
```diff
@@ -1,42 +1,93 @@
...
+    while (index < expression.size() && isdigit(expression[index])) {
+      num_str.push_back(expression[index]);
+      index++;
+    }
+  }
+  else{
+    while (expression[index]!='}' && index<=expression.size()) {
       assert(expression[index]!='{');
-      assert(index<expression.size());
+      assert(index<=expression.size());
```
DominikLeisinger, line 77: This assert may be redundant, as it is the same requirement as the while loop's exit condition.

Group-05/src/DataGrid.cpp
```diff
@@ -367,4 +468,519 @@ const DataGrid::row_t &DataGrid::at(std::size_t row_index_) const {
...
+  return std::accumulate(double_values.begin(), double_values.end(), 0.0) /
+      static_cast<int>(double_values.size());
+}
+
+/**
+ * Calculates the median of the vector.
+ * @param double_values The given doubles vector.
+ * @return The median of the vector
+ */
+constexpr double DataGrid::calculateMedian(std::vector<double> double_values) const {
```
iamrb1, line 741: I think a possible optimization is to pass double_values as a const reference and then after the basic (if double values.empty()) check, you can then create a copy of the parameter vector to sort using ranges. In this way, you do not need to copy the vector each time you pass it into the function since it may be empty (assuming you do not want to modify double_values). This is only a suggestion; however, since you may not want modify double_values as mentioned.

Group-05/src/ExpressionParser.cpp
```diff
@@ -74,85 +134,195 @@ auto cse::ExpressionParser::MakeSubtractFun(const std::string &key1, const std::
...
-    return c=='+' || c=='-' || c=='/' || c=='*';
-  }));
-  size_t index = 0;
-  std::string first_key;
-  std::string second_key;
-  first_key = ParseKey(expression, index);  
-  double result = 0;
-  assert(symbol_table_.find(first_key)!=symbol_table_.end());
+std::function<double(std::map<std::string,double> &)> cse::ExpressionParser::MakeFunc( const std::string &expression, int precedence, size_t &index) {
+  std::map<std::string, double> symbol_table;
```
DominikLeisinger, line 220: From what I can tell, this symbol_table variable is only used in code that has been removed, consider removing it for conciseness?

Group-05/src/DataGrid.cpp
iamrb1: Essentially creating a mutable copy before sorting after the if check to not create a copy for the if check as well

Group-05/src/DataGrid.cpp
iamrb1: I agree with this change -> I also pointed it out

Group-05/src/DataGrid.cpp
```diff
@@ -367,4 +468,519 @@ const DataGrid::row_t &DataGrid::at(std::size_t row_index_) const {
   return grid_[row_index_];
 }
 
+/**
+ * Returns a DataGrid of two merged DataGrids by Row or Column
+ * @param grid_two
+ * @param row_append
+ * @return merged DataGrid
+ */
+DataGrid DataGrid::merge(const DataGrid &grid_two, bool row_append) const {
```
iamrb1, line 477: I think an if check can be added within the beginning to first see if grid_two is an empty DataGrid. Performing that if check will save the overhead of copying the DataGrid and performing merge operations. Lastly, I think that the code functionality can be written to perform less copies instead of copying every single time. I would suggest trying to rewrite the functionality if possible to merge the DataGrids into an empty DataGrid rather than copying over everything from the initial datagrid. If this is not possible, disregard this altogether.

Group-05/src/ExpressionParser.cpp
```diff
@@ -74,85 +134,195 @@ auto cse::ExpressionParser::MakeSubtractFun(const std::string &key1, const std::
...
-  assert(std::any_of(expression.begin(), expression.end(), [](char c) {
-    return c=='+' || c=='-' || c=='/' || c=='*';
-  }));
-  size_t index = 0;
-  std::string first_key;
-  std::string second_key;
-  first_key = ParseKey(expression, index);  
-  double result = 0;
-  assert(symbol_table_.find(first_key)!=symbol_table_.end());
+std::function<double(std::map<std::string,double> &)> cse::ExpressionParser::MakeFunc( const std::string &expression, int precedence, size_t &index) {
```
DominikLeisinger, line 219: This function is very complex, consider either splitting it up into other functions or having inline comments with a quick description of what is happening?

Group-05/src/ExpressionParser.cpp
```diff
@@ -1,42 +1,93 @@
...
 /**
  * @brief Parses the next key in the equation, returning that number as a double
  * 
  * @param expression The string representing the equation
  * @param index The position of the parser in the string
  * @return The first key found
  */
-std::string cse::ExpressionParser::ParseKey(const std::string &expression, size_t &index) {
-  assert(index < expression.size());
+std::function<double(std::map<std::string,double> &)> cse::ExpressionParser::ParseKey(const std::string &expression, size_t &index) {
```
DominikLeisinger, line 43: Another decently complex function, could use some line comments

iamrb1 (commented):
### DataGrid
### Design

1. **Informative identifiers** -> Names of variables, functions, types, etc are well described and easy to understand in Datagrid. For example, the underlying data structure supporting the DataGrid is a vector of a vector of Datums accurately named “grid_”. Additionally, all of the methods of DataGrid give a clear scope on functionality due to a concise name such as deleteRow, insertRow, columnLessThan etc (there are no ambiguous or verbose method names that I can find within DataGrid). Aliases are used (such as row_t for std::vector<Datum>). This promotes simple readability for anyone to understand the code. The enum class operations contained standard operation names that specified functionality (for example, LESS_THAN). Overall, the entirety of the code followed suit on making names simple and straightforward to promote readability.
2. **Avoids “magic numbers”** -> The code avoids “magic numbers” really well, rather utilizing std::numeric_limits<std::size_t>::max() for default values of indices. With std::numeric_limits being constexpr and repeated use of this functionality, you could put a static constexpr MAX constant variable at the top of the class or some other identifier to replace this code and remove redundancy since it always gives the max size of a size_t. The code additionally utilizes 0 and empty strings which do not have to be set within a constant variable as they are not considered “magic numbers” overall. Great job avoiding magic numbers nonetheless!
3. **Intuitive and functional interface** -> DataGrid’s underlying data structure is correctly marked as a private member variable as the user should not be able to access it. The enum class operations are also correctly marked under private. There are private methods that are utilized within public methods to expose the necessary functionality only. The struct DataGridMathSummary is correctly marked under public to allow the user to access the struct when calling public methods such as dataGridMathSummary in which standard math operations such as mean, median, mode etc are given to the user. All the above points listed display the great encapsulation followed within DataGrid. The class is extremely easy to use and is straightforward, with methods to perform math operations upon the grid such as columnMean etc. There are numerous different constructors that are simple to use when building a DataGrid. Getters and setters are added to allow the user to access rows or columns of the grid, and to add to the grid as well. Overall, great work creating a seamless and well-built functional interface. 
4. **Informative comments** -> Each brief provided at the top of the method gives a high level description of what the method does. The briefs written are very concise and descriptive. Great job! When necessary, informative comments are placed in areas that may need a little bit of guidance for the reader in certain methods. For example, insertDefaultRow provides great comments for explaining the overall structure. However, I do want to point out that it may be beneficial to write a comment to explain certain functionality usefulness as mentioned within the rubric. In particular, line 173 can be explained in the sense of design implementation (why are we static_casting to a std::ptrdiff_t and what is the reasoning behind it?) (these are the sort of lines that could be beneficial explaining to the average reader who might not have pre existing knowledge of the code beforehand.) Regardless, the code is commented in places appropriately! 
5. **Consistent style** -> DataGrid’s member variable contains a lowercase format with a trailing underscore. Additionally, camelCase is utilized in methods throughout the class (for example, getDoubleValues). I just wanted to point out that ExpressionParser contains PascalCase for methods so I do recommend that everyone in the group maintains a consistent style. Otherwise, in terms of DataGrid alone, the style choices are consistent and in an informative manner.
6. **Visual appearance** -> I like the structure implemented within DataGrid a lot as I believed it was easy to follow along. Getters and Setters are bundled together with no line spacing to ensure a format of “similar” methods. Column operations are bundled together and line breaks are given before and after the bundle respectively. Insert, delete, calculate etc are all examples of methods in the “bundle” format. This was a unique way of formatting but I liked it as it was very easy to follow. Clang format is used within the code and spacing is used accordingly. Great job!

### Code optimization & flexibility

Overall, it seems to me that the code is very flexible in terms of the number of constructors a DataGrid can be created from, the number of calculations such as mean, median, mode etc can be made, resizing functionality, sorting functionality, comparisons for columns, printing etc. There are methods for every bit of functionality that you can think of. Due to the underlying structure being a container, many of the operations are quite trivial and are already optimized by the author. I do have some suggestions for optimizing the code; however. For example, in calculateMedian, a small optimization is to pass double_values as a const reference and then after the basic (if double values.empty()) check, you can then create a copy of the parameter vector to sort using ranges. In this way, you do not need to copy the vector each time you pass it into the function since it may be empty (assuming you do not want to modify double_values). 

Additionally, a minor optimization in calculateStandardDeviation is to create a constant variable outside of the loop to calculate the mean of double_values and then setting it within the loop for the subtraction you are performing. Right now, you are calling the calculateMean method for each value in the loop which is unnecessary. This fix should provide a speedup for you.

I would also like to point out something I think can be changed within merge. I think an if check can be added within the beginning to first see if grid_two is an empty DataGrid. Performing that if check will save the overhead of copying the DataGrid and performing merge operations. Lastly, I think that the code functionality can be written to perform less copies instead of copying every single time. I would suggest trying to rewrite the functionality if possible to merge the DataGrids into an empty DataGrid rather than copying over everything from the initial datagrid. If this is not possible, disregard this altogether.

Overall, your code is very well written and I am impressed with the intuitiveness and the functionality! Great job!

Group-05/src/ExpressionParser.cpp
DominikLeisinger, line 1: **Implementation Review for ExpressionParser:**
This file is very well organized and very detailed, minus a couple places that may require some additional elaboration for a newbie to the class. I have left some inline review comments for this. Other than that, this is great code. All relevant functions are grouped together and have great Doxy briefs, as well as some had comments throughout them to further elaborate what is being done. I did not see any code that looked elementary or incorrect, as you showed clear knowledge of what we have learned in class as well as others, e.g lambas and std::function. There were places where some code could be consolidated to helper functions, like MakeFunc, (I left an inline review comment for this, I understand as an already complex function with helper functions being called already that it may be redundant or unnecessary to split it up further) but nowhere did I see any unnecessary code or hodgepodge organization within functions. There were many instances of you using const as not to manipulate any important variables, which is great code practice. Overall solid code with modern constructs and mostly easy to understand for someone new to the class. Great Job!

tests/Group-05/DataGridTests.cpp
DominikLeisinger, line 1: **Testing and Error Checking Review for DataGrid:**
I can immediately tell how much work went into this file, so props to you and your team for developing such intensive and elaborate tests. I observed many necessary asserts in the .cpp file, as well as some error throwing that I can see here were tested for with CHECK_THROWS, however I did not see a CHECK_THROWS for each instance of an error throw, for example your getRow() exception can easily be checked with an empty DataGrid in your test case for it. Other than that, your .cpp file demonstrated great error handling and edge case testing, followed by a monstrous test file that is extremely impressive.

tests/Group-05/DataGridTests.cpp
```diff
@@ -1,161 +1,1459 @@
...
+  grid.deleteColumn(0);
+  auto [rows4, cols4] = grid.shape();
+  CHECK(rows4 == 3);
+  CHECK(cols4 == 0);
 }
 
 /**
  * @brief Test accessing row
  */
 TEST_CASE("DataGrid: Get Row", "[access]") {
```
DominikLeisinger, line 338: Your .cpp file throws an error for an empty DataGrid or a Rowindex that is too large, you could consider a CHECK_THROWS here.

tests/Group-05/ExpressionParserTests.cpp
DominikLeisinger, line 1: **Testing and Error Checking Review for ExpressionParser:**
Overall very elaborate tests, the MakeFunc() function is really getting all bases covered here. You test your constexpr, and have tests for dividing by zero and other invalid expressions which is great testing practice. However, I am not observing any checks for incomplete/undefined expressions? For instance, in your first test case if I submitted: 'func = parser.MakeFunc("sin({val4} + )", 0, default_index)', or 'func = parser.MakeFunc("sin({val4} + {madeupvar})", 0, default_index);' what would happen? Does MakeFunc check for this? It looks like ParseKey checks for some of this, but both some line comments for that function and testing for these possiblities in this file would be extremely helpful for error checking your code. Other than this, your test suite is very comprehensive and deserves recognition for your efforts. Amazing job, you tested your whole files functionality extremely well.

tests/Group-05/ExpressionParserTests.cpp
```diff
@@ -3,40 +3,248 @@
...
   cse::ExpressionParser parser;
-  parser.SetSymbolTable("val1", -2);
-  parser.SetSymbolTable("val2", 2);
-  parser.SetSymbolTable("val3", -15);
-  parser.SetSymbolTable("val4", 15);
-  parser.SetSymbolTable("val5", 0);
-  parser.SetSymbolTable("val6", 30);
-  parser.SetSymbolTable("val7", 1);
+  std::map<std::string, double> symbol_table;
+  symbol_table["val1"] = -2;
```
DominikLeisinger, line 50: You could try some much larger/smaller numbers to ensure that your functionality works for them, like 1000/-1000, which would also put your exponent functionality to the test

Group-05/src/ExpressionParser.h
```diff
@@ -53,66 +82,240 @@ class ExpressionParser {
...
   }
 
+  /**
+   * @brief Get the content inside Parenthesis
+   * 
+   * @param expression string representing equation
+   * @param index position in string
+   * @return std::string content inside parenthesis
+   */
+  std::string GetParenthContent(const std::string& expression, size_t index);
```
iamrb1, line 92: Although the name is perfectly fine as it is easy to understand, I do think changing it to GetParenthesisContent is not too verbose and is much clearer for the average reader.

Group-05/src/ExpressionParser.h
```diff
@@ -53,66 +82,240 @@ class ExpressionParser {
...
+   * @brief Creates multiplication function that take a map with values to add
    *
-   * @param name1 Key for first value
-   * @param name2 Key for second value
+   * @param left Lambda which returns the left value
+   * @param right Lambda which returns the right value
    * @return auto Lambda function to perform operation
    */
-   auto MakeDivideFun(const std::string &name1, const std::string &name2);
+  std::function<double(std::map<std::string,double> &)> MakeDivFun(std::function<double(std::map<std::string,double> &left)>, std::function<double(std::map<std::string,double> &right)>);
```
iamrb1, line 111: Like GetParenthContent, the name captures what the function does quite well; however, it is not too verbose to just say MakeDivFunction instead of Fun. Consider changing it to be more clear to the average reader.

Group-05/src/ExpressionParser.h
```diff
@@ -26,6 +36,15 @@ class ExpressionParser {
...
   }
 
+  /**
+   * @brief Get the Token object
+   * 
+   * @param expression string representing equation
+   * @param index representing position in expression
+   * @return size_t index of next token
+   */
+  size_t getToken(const std::string& expression, size_t &index);
```
iamrb1, line 46: The style for this method (albeit private) is in camelCase when the rest of the (albeit public) methods are in PascalCase. If this is a design choice, ignore this change.

Group-05/src/ExpressionParser.h
```diff
@@ -44,6 +63,16 @@ class ExpressionParser {
     div_char='/';
   }
 
+    /**
+   * @brief set char to division operator
+   * 
+   * @param div_char comparison char
+   */
+  constexpr void exponent(char& exp_char){
```
iamrb1, line 71: These methods set chars to the division operator. I mentioned in the general review as possible implementation to improve flexibility within your code.  I am thinking that these could be potentially changed to be a constexpr getOperator function that contains all of these operators in a switch statement rather than having separate functions or even just constexpr chars that are equal to the operator (ex: constexpr char ADDITION = “+”). See if this is possible to implement, I believe that avoiding the function call (when invoking it within MakeFunc) should give a speedup.

Group-05/src/ExpressionParser.h
```diff
@@ -53,66 +82,240 @@ class ExpressionParser {
...
+   * @brief Creates multiplication function that take a map with values to add
    *
-   * @param name1 Key for first value
-   * @param name2 Key for second value
+   * @param left Lambda which returns the left value
+   * @param right Lambda which returns the right value
    * @return auto Lambda function to perform operation
    */
-   auto MakeAddFun(const std::string &name1, const std::string &name2);
+  std::function<double(std::map<std::string,double> &)> MakeAddFun(std::function<double(std::map<std::string,double> &left)>, std::function<double(std::map<std::string,double> &right)>);
```
iamrb1, line 121: I really like that these helper functions are marked under private. This is a good design implementation when the user should only be able to access MakeFunc. Getters and Setters for the member variable are also added. Encapsulation is highly prioritized well. Great job!

Group-05/src/ExpressionParser.cpp
iamrb1: I agree with Dom here; I think that ParseKey should contain insightful comments to explain certain design choices and to assure the reader on what the code's functionality consists of.

Group-05/src/ExpressionParser.cpp
iamrb1, line 219: Along with ParseKey, this function should be detailed with comments to guide the reader on how the functionality works due to its complexity.

Group-05/src/ExpressionParser.cpp
iamrb1: I also suggest removing the symbol table and its assigned values below. It seems that this was tested in main and was supposed to be removed. There are many cases of left over code that have not been removed. Make sure to clean it up before it being sent to grading.

Group-05/src/ExpressionParser.h
```diff
@@ -53,66 +82,240 @@ class ExpressionParser {
...
+
   /**
    * @brief Evaluates equation represented by expression and returns simplified value as a double
    *
    * @param number_map Map with value and key that will be regerences in expression
    * @param expression Representing equation
    * @return double representing value of equation
    */
-   double Evaluate(const std::string &expression);
+  double Evaluate(const std::string &expression, int precedence, size_t &index);
```
iamrb1, line 309: This seems to be a leftover function. From what I gather is that this used to be the old way of doing MakeFunc. In the diff, it seems that the function is removed in the cpp. Remove this for clarity.

Group-05/src/ExpressionParser.h
```diff
@@ -53,66 +82,240 @@ class ExpressionParser {
...
+
+
+  /**
+   * @brief Creates power function that take a map with values to perform operation
+   *
+   * @param left Lambda which returns the left value
+   * @param right Lambda which returns the right value
+   * @return auto Lambda function to perform operation
+   */
+  std::function<double(std::map<std::string,double> &)> MakeExpoFun(std::function<double(std::map<std::string,double> &left)>, std::function<double(std::map<std::string,double> &right)>);
```
iamrb1, line 175: I mentioned "a singular templated function can be implemented to reduce the number of helper functions in the Make___Fun format" -> See the general review for more details but if this is not possible disregard. I am thinking that this could be done due to similar logic in Expo and Sin.

Group-05/src/ExpressionParser.h
```diff
@@ -53,66 +82,240 @@ class ExpressionParser {
...
+  /**
+   * @brief Get the content inside Parenthesis
+   * 
+   * @param expression string representing equation
+   * @param index position in string
+   * @return std::string content inside parenthesis
+   */
+  std::string GetParenthContent(const std::string& expression, size_t index);
+
   std::map<std::string, double> symbol_table_;
```
iamrb1, line 94: Right now, you are using std::map. I recommend using std::unordered_map if order is not important as it is faster for look up and will optimize your code.

Group-05/src/ExpressionParser.h
```diff
@@ -53,66 +82,240 @@ class ExpressionParser {
...
+/**
+ * @brief Raises a base to an exponent at compile time.
+ *
+ * @tparam T A type that supports multiplication.
+ * @param base The base value.
+ * @param exp The exponent (must be non-negative).
+ * @return The result of base^exp.
+ */
+  template <typename T>
+  static constexpr T constexprPower(T base, int exp) {
```
iamrb1, line 244: I think this function can avoid recursive calls altogether and use a while loop instead. This should optimize by minimizing the overhead that recursive calls invoke.

iamrb1 (commented):
### ExpressionParser

### Design

1.  **Informative identifiers** -> Straightforward variable names such as symbol_table_ and function names such as multiplication, MakeCosFun etc easily explain the functionality of the function while having a concise and short name. It is easy for a reader to understand the code with no prior knowledge regarding it. 
I do think that the functions “Make___Fun” can be renamed to “Make___Function” to provide a consistent naming choice. I do not think it is too verbose if that is a particular concern.
2. **Avoids “magic numbers”** -> The code does not have any use cases where a magic number would be implemented. Due to this reason, there is no need to add constant variables for numbers and thus, the code avoids magic numbers all together.
3. **Intuitive and functional interface** -> The class exposes only the necessary functionality in the sense that all the user needs to do to utilize the expression parser is invoke MakeFunc. This is a public method and MakeFunc is composed of private methods behind the scenes. I do like the design choices to have all the helper functions private. Getters and Setters are provided for the private member variable symbol_table_. Encapsulation is prioritized and it is clearly shown within the code. Nice job! I do think a specific design choice can be changed; however. It seems to me that there are certain private constexpr methods that set an operator string to the string passed in (for example: constexpr void subtraction). I am thinking that these could be potentially changed to be a constexpr getOperator function that contains all of these operators in a switch statement rather than having separate functions or even just constexpr chars that are equal to the operator (ex: constexpr char ADDITION = “+”). See if this is possible to implement, I believe that avoiding the function call should give a speedup.
4. **Informative comments** -> I think that there are areas in the code that desperately need more insightful comments. The functions ParseKey and MakeFunc contain functionality that might not be straightforward to the average reader. Comments throughout those functions could explain certain design choices and what is happening as each if check is hit within. For example, MakeFunc includes comments that simply state “Creates Function”. These comments could be rewritten to describe each comparison check. This would be a great step forward in making your code easy to understand. 
The briefs for each function are informative and explain the overall functionality. You did a great job!
5. **Consistent style** -> It seems to me that a variety of styles are used in your code. For example, both camelCase (ex: getToken) and PascalCase (ex:MakeAddFun) are used for methods. Snake_case is used for member variables, which is excellent. Consider making all methods having the same style throughout. Otherwise, the code is great in terms of style!
6. **Visual appearance** -> I think that the code needs improvement and refinement in terms of the visual appearance. In terms of whitespace, indentation, and line breaks, the code follows a consistent format (I assume clang) and is well written. The only issue I see are lines of code that are unused which are kept. Consider removing certain lines that I will point out to make your code organized. Additionally, I believe the function defined as Evaluate is not present within your code. I think that Make Func is Evaluate so consider removing the definition in your header file.

### Code optimization & flexibility

As mentioned earlier,  It seems to me that there are certain private constexpr methods that set an operator string to the string passed in (for example: constexpr void subtraction). I am thinking that these could be potentially changed to be a constexpr getOperator function that contains all of these operators in a switch statement rather than having separate functions or even just constexpr chars that are equal to the operator (ex: constexpr char ADDITION = “+”). See if this is possible to implement, I believe that avoiding the function call should give a speedup. 

I also think that a singular templated function can be implemented to reduce the number of helper functions in the Make___Fun format. For example, you can instead pass in a parameter to perform which function to make and have default arguments if for example the “right” parameter that is in MakeExpoFun is unused when you want to call MakeCosFun. Essentially see if it is possible to collate the code. 

You should use std::unordered_map instead of std::map for faster look up times. This would be a good change for speedup. 

I noticed that constexprPower utilizes recursion. I think it should be possible to avoid recursive calls by using a while loop in this function. If this change isn’t possible, disregard. However, if it is, I highly suggest doing so. 

Overall, the code is extremely well structured and easy to follow. The code is optimized in other areas and usage of constexpr shows that compile time operations can be done. I think making the templated function that I mentioned earlier should enhance the flexibility of your code; however, it is not necessary as the code performs what the class description asks for. Very impressive code in terms of functionality; great work Group 5!

ze-ntsh:
I will be reviewing the code design and usage of Advanced C++ features in your DataGrid class.
The class shows a solid use of templates, allowing for generic data types, which is a good design choice. 

One key inefficiency in my opinion that I thought could be discussed is the way elements are inserted/the grid being resized. If that sort of operation is frequent, maybe std::vector::reserve could be used before insertion to prevent unnecessary reallocations. 

I also thought the lack of the [] operator overload could be a bit inconvenient but that is purely a design choice decision. 

The constructor could use an initializer list, instead of using the resize method which would eliminate the overhead of the size checking and the other things which are needed in resize.

Loops and nested loops can be replaced with STL algorithms for better performance. I understand that there are many places these are used due to the nature of the class, but an inner helper function can be used. I would also understand if these changes would be too tedious to make.

Apart from this, the class seems to be pretty good. Nice work!

ze-ntsh:
I will now be reviewing the code design and usage of Advanced C++ features in your ExperssionParser class.

The MakeAddFun function and other similar functions contain hardcoded values in a symbol table that aren't used. This seems like debugging code left in production. I would remove these hardcoded values or move them to a proper initialization function.

Functions like addition(), subtraction() etc. (lines 59-97) are marked constexpr but take in a refence to a char as an argument, which just makes the constexpr keyword redundant as references are evaluated during runtime. I could be wrong but let me know if that is the case.

Finally, the MakeFunc/MakeAddFun and other similar function and extremely complex with nested conditions and unclear recursion patterns. I am not sure what these are doing. Some more comments to explain the conditions/base cases would be good.

Lots of use of constants/symbols throughout the code, I understand it, but it could be certainly more dynamic (assign it to member variable or constants).

Honestly, a token-based lexer approach would significantly improve this code.
The current implementation mixes lexical analysis, parsing, and evaluation in a way that's difficult to follow and maintain. The getToken function is a rudimentary attempt at tokenization, but it's inconsistently used and doesn't properly separate concerns.

What I would have done is, create a Token class/struct and a Lexer helper class to convert the expression string into a stream of tokens, then create an abstract syntax tree, which would make this a lot simpler/easy to understand and scale.

Group-05/src/DataGrid.cpp
```diff
@@ -1,37 +1,39 @@
...
 
-/**
- * @brief Index a row of the DataGrid
- * @param row_index_ Index of the row
- * @return Indexed vector row from DataGrid
- */
-DataGrid::row_t &DataGrid::operator[](const std::size_t row_index_) {
-  return this->at(row_index_);
-}
-
```
ze-ntsh, line 27: Maybe the [] operator be reinstated with better bounds checking as for a grid it is quite conventional in my opinion. It could use pre-existing functions like getColumn, getRow to retrieve whole rows and columns or singular values.

Group-05/src/DataGrid.cpp
```diff
@@ -207,21 +237,75 @@ void DataGrid::insertDefaultColumn(std::size_t column_index_,
...
+  // If column_index_ is max, update it to insert at the end
+  if (column_index_ == std::numeric_limits<std::size_t>::max()) {
+    column_index_ = grid_[0].size();
+  }
 
-  if (grid_.empty() || column_index_ > grid_.front().size()) {
+  assert(column_index_ <= grid_[0].size());
+  if (column_index_ > grid_.front().size()) {
     throw std::out_of_range("Column index out of bounds");
   }
```
ze-ntsh, line 254: I have seen this some places but just pointing it out here where I found them together. The asserts and exceptions are a bit "cross-used". You can stick to one instead of using them both for similar purposes.

Group-05/src/DataGrid.cpp
```diff
@@ -367,4 +468,519 @@ const DataGrid::row_t &DataGrid::at(std::size_t row_index_) const {
...
+ * @param value
+ * @return vector of indices
+ */
+std::vector<size_t> DataGrid::search(std::size_t column_index,
+                                     const Datum &value) {
+  std::vector<size_t> result;
+  for (size_t i = 0; i < grid_.size(); ++i) {
+    if (getValue(i, column_index) == value)
+      result.push_back(i);
+  }
```
ze-ntsh, line 614: You guys have a grid so there are lots of for loops which can be changed to STL algorithms. Just pointing one out. Here you could maybe use std::copy_if to achieve the same function.

Group-05/src/DataGrid.cpp
```diff
@@ -367,4 +468,519 @@ const DataGrid::row_t &DataGrid::at(std::size_t row_index_) const {
...
+ * @return The mean, median, mode, standard deviation, min, and max of the data grid
+ */
+DataGrid::DataGridMathSummary DataGrid::dataGridMathSummary() const {
+  std::vector<double> double_values;
+  for (const std::vector<Datum>& row : getDataGrid()) {
+    for (const Datum& value : row) {
+      if (value.IsDouble() && !std::isnan(value.GetDouble())) {
+        double_values.push_back(value.GetDouble());
+      }
+    }
```
ze-ntsh, line 708: One more instance where STL can be used. I am using ChatGPT to generate code that would work in this case. You might need to test this one out. However, you can see the use of std::vector reserve as well which I mentioned in my main comment.

```
std::vector<double> double_values;
const auto& grid = getDataGrid();
double_values.reserve(grid.size() * (grid.empty() ? 0 : grid[0].size()));  // Reserve memory upfront

std::for_each(grid.begin(), grid.end(), [&](const std::vector<Datum>& row) {
    std::copy_if(row.begin(), row.end(), std::back_inserter(double_values),
        [](const Datum& value) { return value.IsDouble() && !std::isnan(value.GetDouble()); });
});
```

Group-05/src/DataGrid.cpp
```diff
@@ -367,4 +468,519 @@ const DataGrid::row_t &DataGrid::at(std::size_t row_index_) const {
...
+ */
+constexpr double DataGrid::calculateMean(const std::vector<double>& double_values) const {
+  if (double_values.empty()) {
+    return std::numeric_limits<double>::quiet_NaN();
+  }
+
+  assert(!double_values.empty());
+  return std::accumulate(double_values.begin(), double_values.end(), 0.0) /
+      static_cast<int>(double_values.size());
+}
```
ze-ntsh, line 734: I believe that this and similar functions aren't really constexpr as you are using a vector as an argument which is dynamically allocated once the program is running but again std::vector was made kinda constexpr in C++20 but not sure if that applies here.

 I don't know if you can switch to array in this case but that could work, or you can simply not mark it constexpr

Group-05/src/DataGrid.h
```diff
@@ -28,24 +29,83 @@ class DataGrid {
...
    * @param default_value_
    */
-  explicit DataGrid(const std::size_t num_rows_ = 0, const std::size_t num_columns_ = 0,
-           const double default_value_ = 0) {
-    this->resize(num_rows_, num_columns_, default_value_);
+  explicit DataGrid(const std::size_t num_rows_ = 0,
+                    const std::size_t num_columns_ = 0,
+                    const double default_value_ = 0) {
+    resize(num_rows_, num_columns_, default_value_);
   }
```
ze-ntsh, line 109: This constructor and the one right below it could use an initializer list to bypass the condition checks in resize and simply assign the values more efficiently. 

Something like this could work (you probably need it to fit it with Datum)

```
explicit DataGrid(std::size_t num_rows_ = 0,
                  std::size_t num_columns_ = 0,
                  double default_value_ = 0)
    : grid_(num_rows_, std::vector<double>(num_columns_, Datum(default_value_))) {}
```

Group-05/src/ExpressionParser.cpp
```diff
@@ -46,10 +97,18 @@ std::string cse::ExpressionParser::ParseKey(const std::string &expression, size_
...
-    return m[key1] + m[key2];
+std::function<double(std::map<std::string,double> &)> cse::ExpressionParser::MakeAddFun(std::function<double(std::map<std::string,double> &)> left, std::function<double(std::map<std::string,double> &)> right) {
+  std::map<std::string, double> symbol_table;
+  symbol_table["val1"] = -2;
+  symbol_table["val2"] = 2;
+  symbol_table["val3"] = -15;
+  symbol_table["val4"] = 15;
+  symbol_table["val5"] = 0;
+  symbol_table["val6"] = 30;
+  symbol_table["val7"] = 1;
```
ze-ntsh, line 108: Are val1... val7 used or they just there for debugging purposes. I would move it out anyway.

Group-05/src/ExpressionParser.cpp
```diff
@@ -74,85 +134,195 @@ auto cse::ExpressionParser::MakeSubtractFun(const std::string &key1, const std::
...
+    if (index>=expression.size()){
+      return l1;
+    }
+    char op = expression.at(index);  
     char comparison_char;
-    if (op=='+' || op=='-' || op=='*' || op=='/') {
-      index++;  
-      second_key = ParseKey(expression, index);
-      assert(symbol_table_.find(second_key)!=symbol_table_.end());
+    if (op=='+' || op=='-' || op=='*' || op=='/' || op=='c' || op=='s' || op=='^' || op=='(') {
```
ze-ntsh, line 241: Consider making a set of operators based on different properties and using contains to check or you can use bitwise flags.

Group-05/src/ExpressionParser.cpp
```diff
@@ -74,85 +134,195 @@ auto cse::ExpressionParser::MakeSubtractFun(const std::string &key1, const std::
...
+        }
+        if (expression.substr(index, 4) == "cos(" || expression.substr(index, 4) == "sin(") {
+          return index;  // Return the index where 'cos(' or 'sin(' starts
+        }
+        
+    }
+    
+    // If no operator is found, return a special value indicating no operator
+    return std::string::npos;  // std::string::npos is the largest possible value for size_t
 }
```
ze-ntsh, line 321: Honestly, I don't like how you are getting tokens. Use regex to group similar stuff. Make token into a class/struct to have a consistent interface to interact with them. It will make the class a lot less convoluted.

Group-05/src/ExpressionParser.h
ze-ntsh: These functions can't be constexpr till they have a char reference as an argument which are evaluated during runtime. Also, yes, they can be grouped into.

Group-05/src/ExpressionParser.h
```diff
@@ -53,66 +82,240 @@ class ExpressionParser {
...
    * @brief Creates multiplication function that take a map with values to add
    *
-   * @param name1 Key for first value
-   * @param name2 Key for second value
+   * @param left Lambda which returns the left value
+   * @param right Lambda which returns the right value
    * @return auto Lambda function to perform operation
    */
-   auto MakeMultiplyFun(const std::string &name1, const std::string &name2);
+  std::function<double(std::map<std::string,double> &)> MakeMulFun(std::function<double(std::map<std::string,double> & left)>, std::function<double(std::map<std::string,double> & right)>);
```
ze-ntsh, line 140: All these make____fun could be grouped together into one with another argument to return the expected output (could be the operator/bitwise flag/Token I suggested before).

Group-05/src/ExpressionParser.cpp
PMitkiewicz: I feel like it's a bit overkill tbh, all we need is the position within the string not the token itself. I'll see what I can do though!

Group-05/src/DataGrid.h
ShahaabAli: Thank you for the suggestion!

Group-05/src/DataGrid.h
ShahaabAli: Thank you for the suggestion! This will make the code more clear and less redundant.

tests/Group-05/DataGridTests.cpp
ShahaabAli: We can't test throws because we are using asserts. the throws are for when the application is used in non-debug mode. Thank you for the suggestion!

tests/Group-05/DataGridTests.cpp
ShahaabAli: We can't test throws because we are using asserts. the throws are for when the application is used in non-debug mode. Thank you for the suggestion!

Group-05/src/DataGrid.cpp
ShahaabAli: Thank you for the great suggestion

Group-05/src/DataGrid.cpp
ShahaabAli: Thank you for the suggestion. More comments especially for ChatGPT related implemenations would be very helpful for code readability. Will add some more comments!

Group-05/src/DataGrid.cpp
ShahaabAli: The previous assert removes this potential error. 

assert(row_index_ < grid_.size());

row_index_ is of type size_t, so it must be greater than or equal to 0. If the grid is empty, its size will be zero, and the assert will fail.

Group-05/src/DataGrid.cpp
ShahaabAli: Thank you for your feedback!

ShahaabAli: Thank you for your feedback!

Group-05/src/DataGrid.cpp
MaxKrawec: Changed functions to PascalCase - Thanks

Group-05/src/DataGrid.cpp
MaxKrawec: The function was moved to the .h file.

Group-05/src/DataGrid.cpp
MaxKrawec: I believe the main purpose of using STL algorithms is to improve readability, reduce complexity, and increase efficiency. Personally, I find the original code more readable and less complex. Although the code you recommend may be more efficient, I believe the benefits of readability and simplicity outweigh that. Although there are many cases where STL algorithms improve readability and reduce complexity, I don't believe that in this case.

Group-05/src/DataGrid.h
MaxKrawec: Thanks!

Group-05/src/DataGrid.h
MaxKrawec: Thanks!

Group-05/src/DataGrid.h
MaxKrawec: Thanks!

Group-05/src/DataGrid.h
MaxKrawec: Thanks!

tests/Group-05/DataGridTests.cpp
MaxKrawec: Thanks!

Group-05/src/DataGrid.cpp
MaxKrawec: The professor reviewed my code and didn’t see any issues with it, so I think it might depend on the compiler. I talked to him after class and he said the current code is okay. The tests pass for me and 3 other team members.

Group-05/src/DataGrid.cpp
MaxKrawec: We used asserts for debug mode and  used exceptions for non-debug mode. We used this strategy for our intial classes, and the professor seemed to be okay with it. He left no comments about the asserts or exceptions.

Group-05/src/DataGrid.cpp
MaxKrawec: We used asserts for debug mode and  used exceptions for non-debug mode. We used this strategy for our intial classes, and the professor seemed to be okay with it. He left no comments about the asserts or exceptions.

Group-05/src/DataGrid.cpp
MaxKrawec: Fixed - Thanks

Group-05/src/DataGrid.cpp
MaxKrawec: I believe they can because, according to cppreference, the vector methods are constexpr. Additionally, I had the professor review this code and he didn't see any issues with it.

Group-05/src/ExpressionParser.h
asifmasood7: Noted!

tests/Group-05/ExpressionParserTests.cpp
asifmasood7: Thanks for that detailed feedback, I agree that adding tests for incomplete/undefined expressions is great way to ensure that we can make our error checking more stronger than what it currently holds. Thanks for pointing that part out.

Group-05/src/ExpressionParser.h
asifmasood7: We chose separate functions for clarity, but I will have to agree that using a single constexpr getOperator (or constexpr constants) could potentially streamline the code. I am going to keep looking into it for future iterations and see if this helps with the speedup. Thank u!

Group-05/src/ExpressionParser.h
greencalen17: Used a while loop instead. Thanks

greencalen17:
> ### ExpressionParser
> ### Design
> 1. **Informative identifiers** -> Straightforward variable names such as symbol_table_ and function names such as multiplication, MakeCosFun etc easily explain the functionality of the function while having a concise and short name. It is easy for a reader to understand the code with no prior knowledge regarding it.
>    I do think that the functions “Make___Fun” can be renamed to “Make___Function” to provide a consistent naming choice. I do not think it is too verbose if that is a particular concern.
> 2. **Avoids “magic numbers”** -> The code does not have any use cases where a magic number would be implemented. Due to this reason, there is no need to add constant variables for numbers and thus, the code avoids magic numbers all together.
> 3. **Intuitive and functional interface** -> The class exposes only the necessary functionality in the sense that all the user needs to do to utilize the expression parser is invoke MakeFunc. This is a public method and MakeFunc is composed of private methods behind the scenes. I do like the design choices to have all the helper functions private. Getters and Setters are provided for the private member variable symbol_table_. Encapsulation is prioritized and it is clearly shown within the code. Nice job! I do think a specific design choice can be changed; however. It seems to me that there are certain private constexpr methods that set an operator string to the string passed in (for example: constexpr void subtraction). I am thinking that these could be potentially changed to be a constexpr getOperator function that contains all of these operators in a switch statement rather than having separate functions or even just constexpr chars that are equal to the operator (ex: constexpr char ADDITION = “+”). See if this is possible to implement, I believe that avoiding the function call should give a speedup.
> 4. **Informative comments** -> I think that there are areas in the code that desperately need more insightful comments. The functions ParseKey and MakeFunc contain functionality that might not be straightforward to the average reader. Comments throughout those functions could explain certain design choices and what is happening as each if check is hit within. For example, MakeFunc includes comments that simply state “Creates Function”. These comments could be rewritten to describe each comparison check. This would be a great step forward in making your code easy to understand.
>    The briefs for each function are informative and explain the overall functionality. You did a great job!
> 5. **Consistent style** -> It seems to me that a variety of styles are used in your code. For example, both camelCase (ex: getToken) and PascalCase (ex:MakeAddFun) are used for methods. Snake_case is used for member variables, which is excellent. Consider making all methods having the same style throughout. Otherwise, the code is great in terms of style!
> 6. **Visual appearance** -> I think that the code needs improvement and refinement in terms of the visual appearance. In terms of whitespace, indentation, and line breaks, the code follows a consistent format (I assume clang) and is well written. The only issue I see are lines of code that are unused which are kept. Consider removing certain lines that I will point out to make your code organized. Additionally, I believe the function defined as Evaluate is not present within your code. I think that Make Func is Evaluate so consider removing the definition in your header file.
> 
> ### Code optimization & flexibility
> As mentioned earlier, It seems to me that there are certain private constexpr methods that set an operator string to the string passed in (for example: constexpr void subtraction). I am thinking that these could be potentially changed to be a constexpr getOperator function that contains all of these operators in a switch statement rather than having separate functions or even just constexpr chars that are equal to the operator (ex: constexpr char ADDITION = “+”). See if this is possible to implement, I believe that avoiding the function call should give a speedup.
> 
> I also think that a singular templated function can be implemented to reduce the number of helper functions in the Make___Fun format. For example, you can instead pass in a parameter to perform which function to make and have default arguments if for example the “right” parameter that is in MakeExpoFun is unused when you want to call MakeCosFun. Essentially see if it is possible to collate the code.
> 
> You should use std::unordered_map instead of std::map for faster look up times. This would be a good change for speedup.
> 
> I noticed that constexprPower utilizes recursion. I think it should be possible to avoid recursive calls by using a while loop in this function. If this change isn’t possible, disregard. However, if it is, I highly suggest doing so.
> 
> Overall, the code is extremely well structured and easy to follow. The code is optimized in other areas and usage of constexpr shows that compile time operations can be done. I think making the templated function that I mentioned earlier should enhance the flexibility of your code; however, it is not necessary as the code performs what the class description asks for. Very impressive code in terms of functionality; great work Group 5!

Group-05/src/ExpressionParser.h
MaxKrawec: Fixed

Group-05/src/ExpressionParser.h
MaxKrawec: Fixed

Group-05/src/ExpressionParser.cpp
MaxKrawec: Added comments

MaxKrawec: Added comments

Group-05/src/ExpressionParser.h
MaxKrawec: I believe the professor gave this name as an example (MakeAddFun). That's why we did named it like that.

Group-05/src/ExpressionParser.h
MaxKrawec: Will look into this, especially for the main application. Thank you!

Group-05/src/ExpressionParser.cpp
MaxKrawec: Thanks - updated the code with some of your suggestions!

MaxKrawec: Thanks - updated the code with some of your suggestions!

MaxKrawec: Thanks - updated the code with some of your suggestions!

Group-05/src/ExpressionParser.cpp
MaxKrawec: Removed

Group-05/src/ExpressionParser.cpp
MaxKrawec: Removed

Group-05/src/ExpressionParser.cpp
MaxKrawec: Added comments

Group-05/src/ExpressionParser.h
MaxKrawec: Removed

Group-05/src/ExpressionParser.cpp
MaxKrawec: Removed

tests/Group-05/ExpressionParserTests.cpp
MaxKrawec: Since it's arithmetic, it should work the same. However, this would be good to test for the future. Thanks!

Group-05/src/ExpressionParser.cpp
MaxKrawec: Changed.

Group-05/src/ExpressionParser.h
MaxKrawec: Changed.

Group-05/src/DataGrid.cpp
mercere99: Indeed.. `vector::empty()` is supposed to be `constexpr` since C++20, account to https://en.cppreference.com/w/cpp/container/vector/empty

Some older compilers might be a bit behind in the spec?  If you figured out the problem, please send me the exact error and run your compiler with the `--version` flag and I'm sure we can tease it apart.

mercere99 (approved):
Nice job with `DataGrid` and `ExpressionParser`.  I'm especially happy with the latter given that you don't have previous compilers experience.  Very good interaction with reviewers.  I have a handful of inline comments in the code with the one important one being that, as the reviewers mentioned, you shouldn't mix asserts and exceptions for the same error (more details inline as to why).  My apologies if I missed that on the previous review.

Overall, though, well done!

tests/Group-05/ExpressionParserTests.cpp
MaxKrawec: Thanks

tests/Group-05/DatumTests.cpp
```diff
@@ -18,239 +18,272 @@
...
-      CHECK(std::isnan(string_test.GetDouble()));
-    }
-  }
-
-  SECTION("When the Datum is a double") {
-    cse::Datum double_test(123.123);
-    double_test.AsDouble();
-    CHECK(double_test.GetDouble() == 123.123);
-  }
+  // When the Datum is a string
```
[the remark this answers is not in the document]
MaxKrawec: Thanks!

tests/Group-05/DatumTests.cpp
```diff
@@ -18,239 +18,272 @@
...
 
 TEST_CASE("GetString() functionality", "[datum]") {
-  SECTION("When Datum contains a string") {
-    cse::Datum string_test("test");
-    CHECK(string_test.GetString() == "test");
-  }
-
-  SECTION("When Datum contains a double") {
-    cse::Datum double_test(123.123);
-    CHECK_THROWS(double_test.GetString());
```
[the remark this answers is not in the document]
MaxKrawec: I refactored the code to hit an assert rather than an error when this occurs. Thus, I removed this test case

Group-05/src/CSVfile.cpp
```diff
@@ -67,6 +67,8 @@ std::string CSVFile::SanitizeCsvField(const std::string &input, char delimiter)
     std::string result = "\"";
     for (char character : input) {
       // CITE: Used ChatGPT to help develop this logic for escaping double quotes in CSV fields.
+      // REF: According to CSV standard (RFC 4180) and resaerch on it, double quotes in a field should be escaped by doubling them.
```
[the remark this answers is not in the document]
MaxKrawec: Fixed

tests/Group-05/DataGridTests.cpp
MaxKrawec: Thanks!

MaxKrawec:
This is an old PR for our advanced classes.

We created a new PR from our main application branch --> main. Please review and use that one. However, this PR contains the code reviews for the second checkpoint.

The new PR: https://github.com/CSE498/CSE498-Spring2025/pull/256

PR #188, opened 2025-03-27
Group 09 Liskov Advanced

owenhaiar:


Group-09/Event/Event.hpp
rahnsieg, line 1: **Design:**
Informative Identifiers:
- The identifiers are all fine and there are no issues
Magic Numbers:
- No use of magic numbers
Intuitive and functional interface:
- Everything is intuitive and easy to read
Informative comments:
- I would potentially add a comment explaining the main use case of the event class
- Otherwise everything looks great
Consistent style:
- No issues with style, the comments and the naming conventions  seem to stay the same throughout
Visual Appearance:
- Visually clear with no issues throughout
**Use of advanced C++ features:**
- Entire class is templated and templated correctly
- Uses std::forward and std::apply
- Uses no discard which warns the user if the return value is ignored

Group-09/EventManager/EventManager.hpp
rahnsieg, line 1: **Design:**
Informative Identifiers:
  - Identifiers are informative, clear and easy to understand
Magic Numbers:
  - No magic numbers
Intuitive and functional interface:
  - All member variables are private, class is easy to use and only exposes the necessary functionality, obvious functions exist for common tasks.
Informative comments:
  - Comments are informative and helpful
  - Especially liked the brief overview of the class overall
Consistent style:
  - Style is consistent throughout with the same style of comments and code throughout the entire file
Visual Appearance:
  - Code is easy to read with a good amount of comments, whitespace and indentions
**Use of advanced C++ features:**
  - Class is templated and uses a variadic template
  - Although lamdas are not being uses there is support for them with std::function
  - Value semantics are being used 
Overall the class looks great and I have no negative feedback!

Group-09/Event/Event.hpp
```diff
@@ -1,29 +1,63 @@
...
+  /**
+   * @brief Execute the event by calling its function with its arguments
+   */
+  void execute() { std::apply(function_, args_); }
+
+  /**
+   * Equality based on id
+   */
+  bool operator==(const Event &other) const { return id_ == other.id_; }
+  bool operator!=(const Event &other) const { return id_ != other.id_; }
```
matrix373, line 60: Starting with C++20, you do not need to implement != operator, while having it doesn't affect anything.

Group-09/Event/Event.hpp
```diff
@@ -1,29 +1,63 @@
...
+        args_(std::make_tuple(std::forward<Args>(args)...)) {}
+
+  /**
+   * @brief Get the id of the event
+   */
+  [[nodiscard]] int getID() const { return id_; }
+
+  /**
+   * @brief Get the time of the event
+   */
```
matrix373, line 40: Might want to include noexcept for getID and getTime since these functions will not throw an exception.

Group-09/Event/Event.hpp
```diff
@@ -1,29 +1,63 @@
...
+  [[nodiscard]] int getTime() const { return time_; }
 
-} // namespace cse
+  /**
+   * @brief Set the time of the event
+   * @param time The new time for the event
+   * @attention This will not re-heapify an EventQueue, use EventQueue::update()
+   * for that
+   */
+  void setTime(int time) { time_ = time; }
```
matrix373, line 49: You can also have noexcept and constexpr for this function too.

Group-09/Event/Event.hpp
matrix373: You can also include noexcept and constexpr for == operator.

Group-09/Event/Event.hpp
matrix373: If you are not using Event class in the runtime specifically for the application, you can mark getId and getTime constexpr.

matrix373 (commented):
**Advanced C++ Features**
Templates are used for the class in a meaningful way. Specifically variadic templates are used.
C++ Attributes used properly when necessary.

**Impelemtation Quality**
No warnings encountered with -Wall -Wextra -Werror.
Use of namespace cse is proper.
Code is modular as best as it can.
C++ modern constructors are used as Initializers list for the constructor. 
Memory management handled properly and there is no memory leaks.
Const correctness is used when it can be properly.
Proper and good use of templates and lambda.
Some of the functions can be marked noexcpet, also constexpr if and only if you will not use them in runtime.

Overall great job.

Group-09/EventManager/EventManager.hpp
```diff
@@ -1,48 +1,188 @@
...
+   * @param args The arguments to be passed to the function
+   * @return added Event
+   */
+  std::optional<Event<Args...>> AddEvent(int time,
+                                         std::function<void(Args...)> func,
+                                         Args... args) {
+    if (time < 0) {
+      throw std::invalid_argument("Time must be positive");
+    }
+    Event<Args...> event(next_id_, time, func, std::forward<Args>(args)...);
```
matrix373, line 120: std::move can be used for func parameter, this should avoid copying it.

Group-09/EventManager/EventManager.hpp
```diff
@@ -1,48 +1,188 @@
...
+  bool ResumeEvent(Event<Args...> &event) {
+    int event_id = event.getID();
+    assert(((paused_events_.find(event_id) != paused_events_.end()) ||
+            running_events_.count(event_id)) &&
+           "Event ID must be a managed ID.");
+    if (running_events_.count(event_id)) {
+      return true;
+    } else if (paused_events_.find(event_id) != paused_events_.end()) {
+      if (event.getTime() <= this->getTime()) {
+        event_queue_.add(paused_events_.at(event_id));
```
matrix373, line 162: std::move might be used here to avoid copy

Group-09/EventManager/EventManager.hpp
```diff
@@ -53,15 +193,37 @@ class EventManager {
    * @brief Get the number of events being managed
    * @return The number of events in event_queue_ as size_t
    */
-  size_t getNumEvents() const { return (paused_events_.size() + running_events_.size()); }
+  size_t getNumEvents() const {
```
matrix373, line 196: noexcept can be used for this function

Group-09/EventManager/EventManager.hpp
```diff
@@ -53,15 +193,37 @@ class EventManager {
...
+    return (paused_events_.size() + running_events_.size());
+  }
 
   /**
    * @brief Get the current internal time
    * @return The current time in seconds
    */
-  int getTime() const { return clock_time_; }
+  double getTime() {
```
matrix373, line 204: This function can be const, noexcept, and constexpr

Group-09/EventManager/EventManager.hpp
```diff
@@ -53,15 +193,37 @@ class EventManager {
...
+            .count();
+    return nanoSeconds / std::pow(10, 9);
+  }
 
-};
+  /**
+   * @brief Get the set of running events
+   * @return A copy of the running events set
+   */
+  std::set<int> getRunningEvents() const { return running_events_; };
```
matrix373, line 218: This function can be marked noexcept and constexpr

Group-09/EventManager/EventManager.hpp
```diff
@@ -53,15 +193,37 @@ class EventManager {
...
 
-}
+  /**
+   * @brief Get the map of paused Events
+   * @return A copy of the paused events map
+   */
+  std::unordered_map<int, Event<Args...>> getPausedEvents() const {
```
matrix373, line 224: This function can be marked noexcept and constexpr

Group-09/EventManager/EventManager.hpp
```diff
@@ -1,48 +1,188 @@
...
+
+  /**
+   * @brief Starts queue allowing for time updates
+   */
+  void StartQueue() {
+    if (!running_) {
+      running_ = true;
+      start_time_ = std::chrono::steady_clock::now();
+    }
+  }
```
matrix373, line 96: These functions can be marked noexcept safely.

matrix373 (commented):
**Advanced C++ Features**
Class is templated, templates are used meaningfully. Specifically, variadic templates.
std::forward use properly when necessary.
std::function used effectively and appropriately with type erasure.

**Implementation Quality**
No warnings encountered with -Wall -Werror -Wextra.
Cse namespace is used properly.
Functions are modular and well separated.
Modern C++ constructs used properly such as std::optional, variadic templates, std::function. and std::froward.
Memory handled properly, no leaks.
In some functions usage of noexcept, const, and constexpr can be added. 
There might be constexpr usage improved for some functions, however, that depends on the application you are developing if it won't use runtime operations.

Overall great job.

Group-09/Event/Event.hpp
aydinorh, line 1: Testing and Error Checking:
Unit Testing - I can see that you are checking for getter functions and execute function. However you are not checking for setTime and operators even though they are very basic functions, I think it would be helpful to test them as well.

Edge Cases - I think it would be nice to see some test cases where you are trying to break your code. For example, you could try to call your event with missing arguments or trying to set your globInt and globStr to different types. 

Assertion - There is not assertion to check for

Error Handling - You are not checking for empty function call. Since you are passing a function to your event, there should be an error handling if the function itself is empty since it is not executable. 

Code Optimization and Flexibility
Your code is pretty straight forward and has no issues of optimization and flexibility as far as I can see.

Group-09/EventManager/EventManager.hpp
aydinorh, line 1: Testing and Error Checking

Unit Testing - Tests check for all of the public functions that are callable.

Edge Cases - Although you throw an exception, I think you should also check for adding an event with negative time. You could try to pause and resume events that are not in queue even though you are returning false when that happens.

Assertions - You are using assertion for pausing, resuming, and repeating for simply guarding against incorrect use which is a good practice. 

Error Handling - Your AddEvent throws a std::invalid_argument when time is negative. You could also use for pause, resume, repeat functions but I think assertion is fine as well. 


Code Optimization and Flexibility

I think your code is very well optimized, however you could use error handling instead of assertion in some parts of the code. Also, using templates adds more flexibility. Overall I don't see any issues regarding optimization and flexibility.

Group-09/Event/Event.hpp
IvanBega, line 1: Event. Code Design, Code optimization & flexibility:


The design of this class is very simple and straightforward. It contains constructors, getters, a setter, and an execute function which calls std::apply. Though the class is small, some advanced and modern C++ features like variadic template and attributes. The use of std::forward makes the code flexible by preserving the value category.


Tests demonstrate the flexibility of a class by creating events with different parameters. Side effects are also tested using a global variable. One thing I recommend is adding an edge case like Event<void>


Overall, I didn’t find any issues with this advanced version of a class

tests/Group-09/TestEventManager.cpp
```diff
@@ -4,129 +4,238 @@
...
+{
+  EventManager<int> eventManager;
+  CHECK(eventManager.getNumEvents() == 0);
+  CHECK(eventManager.getNumPaused() == 0);
+  CHECK(eventManager.getTime() == 0);
+}
+
+TEST_CASE("Check getTime in EventManager", "[EventManager]")
+{
+  EventManager<int> eM;
```
IvanBega, line 27: You can add a test with a more complex type inside EventManager

Group-09/EventManager/EventManager.hpp
IvanBega, line 1: EventManager. Code Design, Code optimization & flexibility:

Similarly to Event, multiple modern C++ features were used for the design, such as variadic template and std::optional. Code is well-designed, no magic number, comments are present where appropriate. Maybe one could add one-two empty lines to ResumeEvent/TriggerEvents/RestartQueue to improve readability.
If multiple CPU-heavy events have the same execution time, they will be having difficulty executing at the same time as there is no multithreading. Maybe you can add a way for the user to check the actual time when the function was executed.

Tests also look good and cover all interface. Suggestion for additional test: if multiple events have the same execution time, ensure they are executed in the correct order.

PraiseHeavyLobster:
Implementation quality
-No warnings on compile time
-Proper use of the cse namespace
-I do not see any use of undefined behavior
-Proper use of modularity
-standard library algorithms are used, such as td::apply, std::make_tuple, and std::forward
-memory is handled correctly
-some members, such as getTime at line 204, could be marked const.

Testing & Error Checking
-tests check for all callable functions.
-testing covers a decent amount of edge cases.
-assert does not seem to be used or needed
-tests do not seem to check for any empty functions.

Overall, you did a very good job!

Group-09/Event/Event.hpp
owenhaiar: Events are created at runtime, but I did mark them noexcept

Group-09/Event/Event.hpp
```diff
@@ -1,29 +1,63 @@
...
+
```
mercere99, line 50: If you are going with the FLEXIBLE approach, you probably want a setArgs function too that would allow you to replace all of the arguments.

mercere99 (commented):
Nice job pulling these classes together.

I have some major inline comments about you use of the Args... template parameter.  The way you have it set up, you're not getting any advantage of saving the args separately from the function to call.  You should either combine them immediate, which would allow you to get rid of the template parameters from the classes (which I dub the "SIMPLE" approach below) or you should give more flexibility where you can override the arguments at any time (the "FLEXIBLE" approach).

I also saw that there was no interaction with reviewers in `EventManager`?  None of the helpful comments are acted on either.

I'm holding off on approving this review because I saw that you removed `.clang-format` from the Group-02 directory.  Was this causing problems or was it unintentional?  I'm trying to make sure that groups don't mess with each other's directory, but if it's causing a problem, that would be an overriding concern.

tests/Group-09/TestEventManager.cpp
```diff
@@ -4,129 +4,248 @@
  */
```
PMitkiewicz, line 5: Overall, great job! The tests are extremely comprehensive, covering most test cases. However, you can always be more extensive. Here are some recmmendations for test cases I don't believe you've covered:

Pausing or resuming non existent events. If you're trying to pause or resume an event that isn't running, what happens?

Triggering when no event exists. What happens if some sort of null or invalid event is used?

What happens if an event is being sued multipe times? Are events mutable? Can events be shared?

Is stopqueue and startqueue idempotent? In other words, what's the behavior if they're run multiple times in succession?

Furthermore, like in your other tests it seems like you're creating the events again each time. Try to abstract them, maybe with a function to create the test.

Your use of sleeping each thread may not be good practice. Maybe you could simulate time advancement while you're testng?

Templated versus not templated tests should be split. Other than that, great job!

tests/Group-09/TestApplication.cpp
```diff
@@ -0,0 +1,225 @@
+/**
```
PMitkiewicz, line 1: Like I mentioned with TestEvents, great job! The tests are very comprehensive and cover a large variety of edge cases.

One addition I think will improve readability and efficiency, try abstracting the initiation for each test case. You can make a function or just initiate it once. You can possibly use SECTIONS within a single test case so you only have to initiate once.

Maybe try testing beyond the boundaries. What happens if I call GoTo(-1). What about empty URLS or strings?

Is anything in this application time sensitive? If so, you should make sure its not lagging past whatever your tolerance is for time.

Maybe expand stop testing along with start.

Other than that, these tests are great. I specifically like how separated everything is. Great job!

tests/Group-09/MockClasses/MockImage.cpp
```diff
@@ -0,0 +1,162 @@
+/**
+ * @file Image.cpp
```
PMitkiewicz, line 2: I very much like the use of these mocks for testing, they're very good practice.

PMitkiewicz (commented):
Overall, for the tests, you did a great job. Check the event and application tests for specifics. For general advice, while the tests are very robust, you should test more junk input like negative numbers or empty urls. Also, you can try using section within one test case so that you don't have to initialize stuff for the tests over and over. Great job!

Group-09/EventManager/EventManager.hpp
```diff
@@ -1,48 +1,194 @@
...
+
+  /**
+   * @brief Adds event to paused_events
+   * @param event The event to be paused
+   * @return true if successful, false if event does not exist in queue
+   */
+  bool PauseEvent(const Event<Args...> &event) {
+    int id = event.getID();
+    assert(((paused_events_.find(id) != paused_events_.end()) ||
```
rodriguezjoanna, line 140: Good, but this is guarded only by assert and in release builds the invalid-ID path is UB. I would suggest to either return a status+log or throw std::out_of_range.

Group-09/EventManager/EventManager.hpp
```diff
@@ -1,48 +1,194 @@
...
+    return false;
+  }
+
+  /**
+   * @brief Removes event to paused_events set
+   * @param event The event to be resumed
+   * @return true if successfully removed, false if event is not currently
+   * paused
+   */
+  bool ResumeEvent(Event<Args...> &event) {
```
rodriguezjoanna, line 159: Similar to PauseEvent, this is only guarded with an assert. Resume on an invalid ID is silently ignored in release mode.

Group-09/EventQueue/EventQueue.hpp
```diff
@@ -2,42 +2,158 @@
...
+
+    assert(!removed.has_value() || removed.value().getID() == id);
+    return removed;
+  }
+
+  /**
+   * @brief Pops the event at the top of the EventQueue
+   * @return The event at the top of the EventQueue
+   */
+  Event<Args...> pop() {
```
rodriguezjoanna, line 99: `assert(size()) `is fine for debug, but if this were production code, throwing `std::underflow_error` would recover instead of crashing.

Group-09/Image/Image.cpp
```diff
@@ -58,9 +73,12 @@ Image::Image(const std::string& url, int width, int height, const std::string& a
  * @throws std::invalid_argument If the URL is invalid
  */
 void Image::setURL(const std::string& newURL) {
```
rodriguezjoanna, line 75: Good URL validation Add a unit test for the "ftp://…" case to prove rejection.

Group-09/Image/Image.cpp
```diff
@@ -74,18 +92,17 @@ void Image::setURL(const std::string& newURL) {
  * @throws std::invalid_argument If new dimensions aren' t positive
  */
 void Image::resize(int newWidth, int newHeight, bool maintainAspect) {
```
rodriguezjoanna, line 94: Edge-case test missing of `maintainAspect=true`, `newWidth < newHeight`. Add unit to prove aspect ratio stays correct both ways.

Group-09/TextBox/TextBox.hpp
```diff
@@ -152,9 +155,9 @@ class TextBox {
    * @param newHeight
    */
   void resize(int newWidth, int newHeight) {
```
rodriguezjoanna, line 157: You clamp negatives to 0 but i would add a test for oversized values (e.g., 10 000 px) to ensure no overflow or layout break.

Group-09/PresentationManager/PresentationEventManager.hpp
```diff
@@ -0,0 +1,263 @@
...
+			// Check queue and execute
+			while (_active_queue.size() && _slide_elapsed_time >= _active_queue.peek().getTime()) {
+				_active_queue.pop().execute();
+			}
+		}
+
+		/**
+		 * Delete a slide from the EventManager
+		 */
+		void deleteSlide(const int slide_num) {
```
rodriguezjoanna, line 188: Potential double-free, the `while `loop pops from `_active_queue` instead of `_queues[slide_num]` after the first pop. Worth a defensive unit test.

Group-09/index.html
```diff
@@ -0,0 +1,457 @@
...
+    .presentation-zone {
+      background-color: white;
+      width: 100vw;
+      height: 93vh; /* adjusted to make room for taller nav */
+      overflow: hidden;
+      box-shadow: inset 0 0 10px rgba(0, 0, 0, 0.1);
+      z-index: 1;
+    }
+
+    .bottom-nav {
```
rodriguezjoanna, line 30: “Swap” and “Next Slide” icons are the same color, affordance could improve if primary action (Next) had accent color while swap stays neutral.

Group-09/WebLayout/EditElement.js
```diff
@@ -0,0 +1,160 @@
...
+
+document.addEventListener("DOMContentLoaded", () => {
+  // True if formatting menu is open and textbox is editable, false otherwise
+  let formattingLocked = false;
+
+  // ID of the currently active textbox
+  let currentTextboxId = null;
+
+  // Open formatting menu and manage blur (deavtication)
+  document.addEventListener("dblclick", function (e) {
```
rodriguezjoanna, line 13: a ux idea i had to suggest is to show a subtle tooltip like  “Double-click to edit text” on hover for first-time users.

Group-09/WebLayout/EditElement.js
```diff
@@ -0,0 +1,160 @@
...
+        return;
+      }
+
+      const imageEl = document.getElementById(imageID);
+      if (!imageEl) {
+        alert("Image not found!");
+        return;
+      }
+
+      imageEl.style.width = `${newWidth}vw`;
```
rodriguezjoanna, line 147: The prompt for ID/width/height is modal and interrupts flow i suggest just having a drag corner-handle like textboxes would feel more consistent.

rodriguezjoanna (commented):
### Testing & Error Checking
**Strengths**
      • Large Catch2 test suite covers most public functions (add, remove, update, copy/move, JSON import/export).
      • Special-character handling, duplicate-ID logic, and exception paths are exercised.
      • Tests use REQUIRE_THROWS_WITH to confirm the right exceptions.

**Areas to improve**
      • Several error paths rely only on assert. In a release build these assertions disappear, so misuse causes undefined behavior.
       – See EventManager::PauseEvent, EventManager::ResumeEvent, EventQueue::pop.
      • Missing edge-case tests:
             – Oversized dimensions (e.g., 10 000 px).
             – Image::resize with maintainAspect=true when width < height.
             – Invalid protocols such as ftp://….
             – Repeat-event combined with Pause/Resume.
      • PresentationEventManager::deleteSlide may pop from the wrong queue after the first element (possible double-free).
      • No test feeds malformed or huge JSON into loadSlideDeckFromJson.

      Recommendations (added within inline comments)
      1.	Replace critical asserts with exceptions (std::out_of_range, std::underflow_error) or return-status checks.
      2.	Add the edge-case unit tests listed above.
      3.	Write a small “bad-input” script that loads invalid JSON and confirms a clean error message.
      4.	Add a defensive test around deleteSlide.

### User Experience
**Strengths**
      • Drag-and-drop, resize handles, and keyboard shortcuts (ESC, Space) feel natural.
      • Bottom navigation is clear; borders hide automatically in presentation mode.
      • “Save Deck”/“Load Deck” JSON flow is one click.
      • Images always include alt-text (good accessibility).

**Areas to improve**
      • First-time users won’t know about double-click edit or right-click animations. Add a brief help overlay or tool-tips.
      • Images resize through a modal prompt, unlike text boxes which resize by dragging. Add corner handles for images for consistency.
      • Bottom bar background (#C1D8FF) plus light text fails contrast guidelines so I would suggest to darken text or lighten bar.
      • Primary action (“Next Slide”) and secondary (“Swap”) share the same color—give “Next” a stronger accent.
      • Deleting objects is irreversible; show a toast “Deleted – Undo” for a few seconds.

Group-09/WebLayout/WebLayout.hpp
```diff
@@ -0,0 +1,105 @@
...
+#include "../TextBox/TextBox.hpp"
+
+namespace cse {
+
+struct TextBoxLayout {
+  std::shared_ptr<TextBox> textBox;
+  int yPos;
+  int xPos;
+
+  TextBoxLayout() : textBox(), xPos(0), yPos(0){};
```
albodell, line 26: maybe consider changing these to constexpr so you can use them in compile time contexts and just for flexibility in the future

albodell (commented):
Modern c++ implemtation:
overall there are many uses of modern c++ features with templates, nodiscard, formats, noexcept, constexpr, etc.
There isnt too much that can be added but if you do want to bolster your code you could consider making more values into constexpr since there is little drawback and it could give you safer and more efficient code in some areas, you could also look into using concepts into your templates. Overall I don't think any of that is completely necessary but could be considered for future implementation.

Group-09/WebLayout/WebLayout.cpp
```diff
@@ -119,58 +205,261 @@ void WebLayout::PushTextBox(const std::string &msg, const int &width, const int
...
+          layoutDiv.style.visibility = "hidden";
+        }
+      },
+      layoutID.c_str());
+}
 
-    // Verify values are valid
-    if ((it->textBox.getHeight() > 0 && it->textBox.getWidth() > 0) && (it->x >= 0 && it->y >= 0)) {
-      PushTextBox(it->textBox.getText(), it->textBox.getWidth(), it->textBox.getHeight(), it->x, it->y);
+void WebLayout::toggleImage(const cse::ImageLayout &image) {
```
CircleShift, line 338: You may want to provide an overload for just the imageID, so then you could call it with just a string.

Group-09/WebLayout/WebLayout.cpp
```diff
@@ -119,58 +205,261 @@ void WebLayout::PushTextBox(const std::string &msg, const int &width, const int
...
+}
+
+
+/**
+ * Converts image attributes into html
+ * @param id : Object id
+ * @param newX : New x value location for object to be set at
+ * @param newY : New y value location for object to be set at
+ */
+void WebLayout::setPosition(std::string id, int newX, int newY) {
```
CircleShift, line 409: May want to consider using a map for the id to image/textbox storage, since you would get better search times in functions like this. (And as far as I know all the IDs are unique)

Group-09/EventManager/EventManager.hpp
```diff
@@ -1,48 +1,194 @@
...
-  std::optional<Event> AddEvent(int time, std::string data);
-  bool RepeatEvent(cse::Event &event, int time_interval);
-  void StopQueue();
-  void StartQueue();
-  void RestartQueue();
+  /**
+   * @brief Checks for and triggers events
+   */
+  void TriggerEvents() {
+    while (event_queue_.size() && event_queue_.peek().getTime() <= getTime()) {  // Events to be popped
```
albodell, line 58: it looks like this loop is running very often from a basic condition and causing more cpu usage from it, try to add a more condition version of this to be less of a weight, maybe use wait between checks to lower its strain

Group-09/PresentationManager/PresentationManager.hpp
```diff
@@ -0,0 +1,792 @@
...
+
+		/**
+		 * Removes a slide
+		 * @param slide
+		 */
+		void deleteSlide(const Slide& slide) {
+
+			const auto it = std::ranges::find(_slide_deck, slide);
+			int slide_num = it - _slide_deck.begin();
+			auto const new_pos = _current_pos != 0 ? _current_pos - 1 : _current_pos;
```
CircleShift, line 107: Technically you should only go back a slide if you're removing a slide *before* your current position, plus the slide may be invalid.  May want to check for both of those cases.

Group-09/EventManager/EventManager.hpp
```diff
@@ -53,15 +199,37 @@ class EventManager {
...
+   * @return A copy of the running events set
+   */
+  std::set<int> getRunningEvents() const noexcept{ return running_events_; };
 
-}
+  /**
+   * @brief Get the map of paused Events
+   * @return A copy of the paused events map
+   */
+  std::unordered_map<int, Event<Args...>> getPausedEvents() const noexcept{
```
albodell, line 230: there are many instances of copying quite a bit of data, here it looks like the entire map is being copied, maybe use more efficient uses of data or use const& for referencing data

Group-09/PresentationManager/PresentationManager.hpp
```diff
@@ -0,0 +1,792 @@
...
+
+			auto deck = json::parse(exportSlideDeckToJson(this));
+			deck["slides"].erase(deck["slides"].begin() + slide_num);
+			loadSlideDeckFromJson(deck.dump(2).c_str(), this);
```
CircleShift, line 111: I'm a little confused why you don't just remove it from the _slide_deck vector

Group-09/WebLayout/WebLayout.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+  TextBoxLayout(std::shared_ptr<TextBox> textBox, int x, int y)
+      : textBox(std::move(textBox)), xPos(x), yPos(y) {}
+
+  bool operator==(const TextBoxLayout &textboxLayout) const {
```
albodell, line 31: it looks like getFormattedText().getText() is being called multiple times per comparison here, maybe try different way of comparing with ids for example

Group-09/PresentationManager/PresentationManager.hpp
```diff
@@ -0,0 +1,792 @@
...
+		 */
+		void goTo(const int slide_num) {
+			std::cout << "Going to slide " << (slide_num + 1) << std::endl;
+
+            // If running presentation and last slide, go to end
+            if(_running && slide_num >= _slide_deck.size()) {
+              std::cout << "END OF PRESENTATION" << std::endl;
+              auto currentLayout = _slide_deck.at(_current_pos);
+              currentLayout->deactivateLayout();
+              toggleEndScreen(true);
```
CircleShift, line 221: May want to exit here so you don't print an error message (unless that is what you want)

albodell (commented):
Overall implementation quality is very good, there are a few places where quite a lot of copies are going on and some optimizations could be made. In general some areas could be made more efficient with a better algorithm but you can always keep improving, as is things looks great.

Group-09/PresentationManager/PresentationManager.hpp
```diff
@@ -0,0 +1,792 @@
...
+
+		/**
+		 * This handles the edge case where a slide transition occurs while there are remaining object events
+		 * The objects still needs to reappear in the layout
+		 */
+		void resetObjects() { _event_manager.resetObjects(); }
+
+		/**
+		 * @brief Get the Current Position of the manager
+		 * @return int current position in the slide vector
```
CircleShift, line 330: Comment is copied from `getCurrentPos`

Group-09/index.html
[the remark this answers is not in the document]
CircleShift: Overall code design is good.  Functions have readable names and make sense in context.  May want to look into a couple edge cases and extraneous code.  Some variables may be renamed for clarity.  Some of my comments are for potential features, which are not critiques but suggestions.

tranbao01:
CODE DESIGN REVIEW
1. Informative Identifiers
Functions' name are very clear and describe exactly what they do.

2.Avoids "Magic Numbers"
Use static int instead of hardcode

3.Intuitive and Functional Interface
Clear division between public and private methods.
All attributes are private (content, id, etc.).

4.Informative Comments
Good header comments at the top of every file.
Doxygen-style (@brief, @param, @return) used consistently.

5.Consistent Style
Snake_case for variables inside functions.
PascalCase for class names.
Functions are all camelCase — consistent.

6.Visual Appearance
Consistent 2-space or 4-space indentation.
Great line breaking (no super long unreadable lines).
Logical ordering of methods (getters first, modifiers second, etc.).

Overall, the code design is very solid

tranbao01:
USE OF MODERN C++ REVIEW

default constructors and noexcept move constructors are correctly used.
static member functions for ID generation.
Use of assert (modern and safer than just trusting parameters).
std::move is correctly used in move constructors.

Overall, this application did use modern c++ correct and efficiently

Group-09/WebLayout/WebLayout.hpp
merryymary: Great catch. This was originally created when we didn't have IDs for each object, so this was definitely outdated!

Group-09/WebLayout/WebLayout.cpp
merryymary: This is a great future implementation to add possibly. Right now, the speed isn't a concern since each web layout has individual images and textboxes, so the size of these vectors is typically small.

Group-09/WebLayout/WebLayout.cpp
```diff
@@ -119,58 +205,261 @@ void WebLayout::PushTextBox(const std::string &msg, const int &width, const int
...
-           // Finds div with presentation-zone id, where all boxes and images will be placed
-           var imageBoxDiv = document.getElementById("presentation-zone");
-           if (imageBoxDiv) {
-             imageBoxDiv.innerHTML +=
-                 "<img src='" + msg + "' style='position: absolute; left: " + x + "vw; top: " + y
-                     + "vh; margin: 0; object-fit: contain; width:" + width + "vw; height:" + height + "vh;' />";
-           }
-         }, url.c_str(), width, height, x, y
-  );
+void const WebLayout::renderImage(const std::string &layoutID,
```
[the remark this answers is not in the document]
merryymary: Technically yes it could be put into this function instead of having getID called in the function call portion. The purpose behind having it just be a parameter is to have the function be purely rendering the image and not doing id validation or calls to information.

Group-09/PresentationManager/PresentationEventManager.hpp
```diff
@@ -0,0 +1,263 @@
...
+			const auto event = Event(target, time, function, _presentation_manager, parseID(target));
+			_queues[origin].add(event);
+			_active_queue = _queues[_current_slide]; // Re-copy the active queue
+		}
+
+		/**
+		 * Delete an event by ID
+		 * @param id
+		 */
+		void removeEvent(const std::string &id) {
```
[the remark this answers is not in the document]
merryymary: Yes you are correctly that events are limited to one per element. This is definitely something we have discussed for future implementation but based on how our code functions right now it is one event per element.  The ideas for rotation and translation are great for future support we could add.

Group-09/TextBox/TextBox.hpp
allienello: Nice catch, added a max value to clamp to

Group-09/index.html
allienello: Great idea, gave swap a neutral color while giving Next a primary color

Group-09/WebLayout/EditElement.js
allienello: Love this idea, added on hover. Since there are no 'users' in our current app, it will just show everytime by default. End goal would definitley be to have it show only for first-time users

Group-09/WebLayout/EditElement.js
allienello: This was one of our to-do list items as well; revised and added it. For sure more consistent

Group-09/PresentationManager/PresentationEventManager.hpp
```diff
@@ -0,0 +1,263 @@
...
+				_id_to_slide.erase(e.getID());
+			}
+			_queues.erase(_queues.begin() + slide_num);
+		}
+
+		/**
+		 * Change context to current slide
+		 * @param slide_num
+		 */
+		void onSlideChanged(const size_t slide_num) {
```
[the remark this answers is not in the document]
merryymary: Typically yes, although in all instances where this function is called there are asserts or checks to ensure this very case. EventManager doesn't have access to which slide exists, its only job is to deal with events of slides, so that's why the error checking is done elsewhere :D

Group-09/Image/Image.hpp
```diff
@@ -0,0 +1,77 @@
+/**
+ * @file Image.hpp
+ *
+ * @author Mia Bodenhorn
+ */
+
+#pragma once
+
+#include <emscripten.h>
```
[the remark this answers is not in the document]
merryymary: In the alert functionality some emscripten functions are used it's just not the typical EM_ASM tag around it so it is easily missed 👍

Group-09/Image/Image.hpp
```diff
@@ -0,0 +1,77 @@
...
+  std::string id;
+  // bool validateURL(const std::string& url) const;
+};
+
+void Alert(const std::string& msg);
+
+#ifdef NDEBUG
+#define em_assert(...)
+#else
+#define em_assert(condition)                                              \
```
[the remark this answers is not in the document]
merryymary: I haven't looked heavily into additional assert functionalities so this could be something we further dive into!

Group-09/index.html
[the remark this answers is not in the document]
merryymary: Thank you for the feedback! In regards to the improvement sections, is there more of a specific on what you mean with can not load slides with images? With all our testing it does reload images, was there a specific use case you had fail?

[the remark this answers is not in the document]
merryymary: Also update: there are changes now so if you add an event or transition again, it will just replace the last event as a way to "change" these!

Group-09/EventQueue/EventQueue.hpp
merryymary: Thank you!

Group-09/index.html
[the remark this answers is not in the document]
CircleShift: Let me check, maybe I can send a file that presents the issue, though you may have fixed it since I did the user-review.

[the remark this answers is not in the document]
CircleShift: [slides.json](https://github.com/user-attachments/files/19986962/slides.json)
This does not load with images.

[slides_no_img.json](https://github.com/user-attachments/files/19986983/slides_no_img.json)
Here is the same json file except I removed the image, which loads.  Might be too big of a image file?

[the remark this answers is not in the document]
merryymary: So the issue was the heigh of the image was saved as 0, so it wasn't loading. We actually fixed this issue yesterday so thank you for the feedback on this!

tests/Group-09/TestEventManager.cpp
graceKC04: Thanks for the feedback! I’ve added some of the suggested test cases.

Group-09/Image/Image.cpp
graceKC04: A unit test case has been added to ensure aspect ratio is working properly.

tests/Group-09/TestApplication.cpp
graceKC04: Additional test cases have been added to TestApplication.cpp

Group-09/EventManager/EventManager.hpp
graceKC04: Thanks for the suggestion! I updated the code to forward fun to avoid unnecessary copies. I kept Args... in the class template for now but understand the benefits of making the method fully variadic.

PR #204, opened 2025-04-06
TextBox editing mode 

allienello:
Textbox content can now be edited when a user double clicks a text box; to save/update, click out of the textbox or hit escape.


miabodenhorn (approved):
sublime

merryymary (approved):

PR #208, opened 2025-04-07
Liskov - fixed duplicate images

merryymary:
Removed loadPage function (rendering was moved into addImage/addTextbox function for easier usage) -> this was causing duplicate images


graceKC04 (approved):

allienello (approved):
heck yeah

PR #228, opened 2025-04-19
Event work

owenhaiar:
Added events for objects. Right click to add an event to a image or textbox

You can now choose timing for slide transitions

Events are saved and loaded in json. Events persist after leaving slide.

Fixed broken json loading and stack overflow for large json files

Fixed esc not working when browser was fullscreen

Small UI cleanups

Increased fps to 60


merryymary (approved):
5 stars

graceKC04 (approved):

PR #230, opened 2025-04-20
Implemented formatting menu with textbox editing

allienello:
You can now edit the color, size, and font of text for an entire textbox. Formatting menu appears when you double click it


miabodenhorn (approved):
Thine latest commit doth bring great joy to mine eyes,
a wondrous feat where now, with but a double click,
the humble textbox reveals its inner self,
laying forth options of hue, measure, and script.

graceKC04 (approved):

PR #232, opened 2025-04-21
Added deleting item functionality (only when dragging)

graceKC04:


allienello (approved):
Let's gooooo

owenhaiar (changes requested):
Error when deleting an item with an associated event, also should be backspace since Mac needs fn + delete to delete

owenhaiar (approved):

merryymary (approved):

PR #240, opened 2025-04-22
[Hamilton/Group 5] Application

MaxKrawec:
To run the application:

1) Navigate to ..\CSE498-Spring2025\Group-05
2) Compile with this command: `g++ -std=c++20 main.cpp FinalApplication.cpp src/Datum.cpp src/CSVFile.cpp src/DataGrid.cpp src/ExpressionParser.cpp src/ReferenceVector.h -o a`
3) Run the application with: ./a or a.exe
4) Use the terminal to navigate through the application

If you are using CLion, you may use the 'Run' button on the UI. 

**_EDIT_**
We created a new PR from our main application branch --> main. Please review and use that one. However, this PR contains the code reviews.

The new PR: https://github.com/CSE498/CSE498-Spring2025/pull/256


Group-05/main.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include <iostream>
+
+#include "FinalApplication.h"
+
+/**
+ * @brief main (Program Entry Point)
+ */
+int main() {
+  FinalApplication final_application;
+  final_application.MainMenu(std::cout, std::cin);
```
Devon-FoxElster, line 18: I like this use of istreams and ostreams to enable testing, my group did something similar.

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1236 @@
...
+        "manipulate, analyze and save CSV file data"
+     << std::endl;
+  os << "Developed by: Max Krawec, Calen Green, Pedro Mitkiewicz, "
+        "Shahaab Ali, and Muhammad Asif Masood"
+     << std::endl;
+  os << std::endl;
+
+  cse::DataGrid grid = GridMenu(os, is);
+
+  while (true) {
```
Devon-FoxElster, line 1191: For this while loop and the other 'infinte' while loops, it may help to add a comment before it specifying what condition will exit the cycle
i.e 
```// User must select q to stop the loop```

File: Group-05/FinalApplication.cpp [hunk 3]
```diff
@@ -0,0 +1,1236 @@
...
+            }
+            grid.InsertColumn(new_column);
+            os << "Column added." << std::endl;
+            // Add a custom column by an equation
+          } else if (method == "e") {
+            std::vector<cse::Datum> new_col;
+            os << "Enter equation. Supported operators +-/*^, indexes in curly "
+                  "braces {}, seperate with a single space: ";
+            std::string equation;
+            is.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
```
Devon-FoxElster, line 941: I think this line makes it more clear what the .ignore function is trying to achieve with the numeric_limits parameter, I recommend using that on lines 1094, 1106, 1118, 1138, and 1150 to avoid using 10,000 as a magic number.

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1236 @@
...
+      ManipulateGridMenu(grid, os, is);
+    } else if (option == "m") {
+      MathMenu(grid, os, is);
+    } else if (option == "c") {
+      ComparisonMenu(grid, os, is);
+    } else if (option == "q") {
+      break;
+    } else {
+      os << "\nInvalid option. Try again." << std::endl;
+    }
```
Devon-FoxElster, line 1232: I liked your use of switch cases in the other submenus, I think it would also be applicable here since all of the options are single characters (and it can compare ASCII values).

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1236 @@
...
+  static constexpr int MAX_DEFAULT_STR_LEN = 100;
+  static constexpr double MAX_DEFAULT_VALUE = 1e20;
+
+  auto trim = [&](const std::string &s) {
+    const auto ws = " \t\r\n";
+    auto first = s.find_first_not_of(ws);
+    if (first == std::string::npos) return std::string{};
+    auto last = s.find_last_not_of(ws);
+    return s.substr(first, last - first + 1);
+  };
```
Devon-FoxElster, line 163: I noticed that this helper lambda exists both in this finction and GridMenu, maybe it would work better as a private variable for the FinalApplication class itself?
I had similar thoughts for the istream and ostream that get passed through every function

Devon-FoxElster:
## Code Design
Overall the use of comments was very good. All functions had clear doxygen comments and inline comments did a good job of outlining each affected area's purpose.

The code itself is logical in its progression through the menus and into helper functions. Your code is very compact visually, which can make it hard to look for specific sections. I like to put some whitespace separation between things like if statements, switches, or just sections of unrelated code to give some visual separation, but that is a personal preference ultimately.

There was a case of magic numbers but I addressed it in an inline comment. 

Naming schemes look consistent to me using snake case for variables and Pascal Case for functions.

Nice job on the application!

Devon-FoxElster:
## User Experience
The interface keeps it nice and simple to enter specific commands, consistently only 1 character or number to choose options and a clear way to back up. 

Some suggestions:
1. I would offer for the interface is the use of colors to highlight the titles and available options(with ANSI color codes or other methods, both Group 6 and Group 10's final application make use of them).
2. Better use of white space to separate the menus/inputs from the outputs of the commands. I noticed this in particular in the print menu, where the printing of the datagrid is directly after the options input and causes it to blend in
3. With a focus on submenus, I think it may be worth a consideration to put the menu path the user is currently on. This is what I mean:
for main menu: `Menu Option`
Grid Manipulation Menu: `Main Menu >> CSV Grid Manipulation Menu`
Print options: `Main Menu >> Grid Manipulation >> Print Options`
That may end up a bit long, so the alternative is to always specify what menu the back command will return to.
4. A personal recommendation would be to move the Print menu up to the main menu above the Edit CSV Grid option, as print being in that submenu wasn't immediately apparent (at first I thought there was no print options). Maybe name the new section "View CSV Grid"?
5. I would also like to see an option on the main menu to import a CSV file even if I imported one earlier, that way if I want to switch which CSV I'm working on at the moment I can do that without restarting the program
6. Additionally, during any multistep data inputs (like when creating a new DataGrid) I think something like 'q' for aborting the command would be very helpful in case the user selected the wrong option
7. I would also recommend that, on the edit page, it communicates what kind of format it expects the new values. (when I selected edit row, it asked me to input the new numbers but it did not say how to properly split them up, like by space or comma.)
   Also, there seems to be a bug related to unexpected values, if I input a letter or incorrect format it seems to start an infinite loop
image


Overall, very nice job! I'm impressed with the range of functionality in the program

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1526 @@
...
+ * @return True if it is a valid custom string, false otherwise.
+ */
+bool FinalApplication::IsValidCustomEquation(const std::string& input, int max_number_value) {
+  // I asked claude.ai to check if a string meets these conditions:
+    // 1) The numbers within the string must be surrounded by {}
+    // 2) The numbers within the {} must be less than a certain value.
+    // 3) The values outside the {} can only be +, -, *, /, and ^
+    // 4) There can't be any other words/values
+  // It generated the regex pattern below, which checks for these requirements
+  std::regex pattern(R"(^(\{(\d+)\}[\s]*([\+\-\*/\^])[\s]*)*\{(\d+)\}$)");
```
LeoSSpecht, line 165: This is a bit out of scope, but if curious you could use a Lexer to do this parsing.
As it knows the regex, and it is not dynamic, it is usually faster to parse.
The professor has one that is quite good.

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1526 @@
...
+    if (first == std::string::npos) return std::string{};
+    auto last = s.find_last_not_of(ws);
+    return s.substr(first, last - first + 1);
+  };
+
+  std::size_t num_rows = 0, num_columns = 0;
+  std::string line, token, type_choice;
+
+  // Read number of rows
+  while (true) {
```
LeoSSpecht, line 222: I think this code could be organized a bit better.
From what I can see the code is supposed to be in a loop while expecting to get an input that is valid.
If not, it shows an error message.
An option would be to create a function that does this, and accepts strings for the message, and functions/lambdas that validate the input. This way the code could be a bit cleaner,

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1526 @@
...
+        os << "\nBelow is the grid you imported: \n";
+        data_grid.Print(os);
+        return data_grid;
+      } catch (const std::exception &e) {
+        os << "Import failed: " << e.what() << "\n" << std::endl;
+        // loop back to menu
+      }
+    } else if (choice == "t") {
+      // Pre-made 5×3 grid
+      std::vector<std::vector<cse::Datum>> premade(5,
```
LeoSSpecht, line 388: Here I believe a factory pattern would have been very useful.
The end functionality would be the same, but the code would be much cleaner, as the premade option could be generated somewhere else.

LeoSSpecht: You could also offer additional premade options, but that would just be a plus!

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1526 @@
...
+    os << "csd: Calculate the standard deviation of a column" << std::endl;
+    os << "cmin: Calculate the min of a column" << std::endl;
+    os << "cmax: Calculate the max of a column" << std::endl;
+    os << "sum: Calculate all statistics for the grid" << std::endl;
+    os << "b: Go back" << std::endl;
+    os << "Enter option: ";
+    std::string option;
+    std::getline(is, option);
+
+    if (option == "cmean") {
```
LeoSSpecht, line 447: Here I would make a different design.
I would make a map of strings to functions.
Each of the options would map to a function of what that option does. This would make clear what each part of the code does, instead of all the "input/output" of the option being on the same function.

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1526 @@
...
+      cse::Datum datum = GetDataValue(os, is);
+
+      os << "\nThe column values: \n";
+      PrintColumn((grid.GetColumn(index)));
+      os << "\n";
+
+      os << "Values equal to the given value:" << std::endl;
+      PrintColumn(grid.ColumnEqual(index, datum));
+      return;
+    } else if (option == "cne") {
```
LeoSSpecht, line 624: Same as the previous comment

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1526 @@
...
+ * @param grid The DataGrid to sort
+ * @param os ostream used for output and testing
+ * @param is istream used for input and testing
+ */
+void FinalApplication::SortSubmenu(cse::DataGrid &grid,
+                                   std::ostream &os = std::cout,
+                                   std::istream &is = std::cin) {
+  int choice = -1;
+  do {
+    os << "\n--- Sort Options ---" << std::endl;
```
LeoSSpecht, line 967: As you do a lot of inputs to get user options, I think it would be a great place to put it in a function that gets the input, and validates it.

tests/Group-05/FinalApplicationTests.cpp
```diff
@@ -0,0 +1,1996 @@
...
+      std::string::npos);
+
+  // Invalid string
+  is.str("Hello!\n0\n");
+  os.clear();
+  final_application.AddSubmenu(grid, os, is);
+  CHECK(os.str().find("Invalid choice. Cannot be a string. Try again.") !=
+      std::string::npos);
+
+}
```
LeoSSpecht, line 1996: Congratulations on the testing, it is obviously impossible to test every possible scenario, but you seem to have done a great job here, testing most inputs, combinations, including invalid ones!

LeoSSpecht:
Congrats on the application!

## Implementation Quality

Overall the code is good, you guys use functions and don't seem to use any bad practices or any topics that would make the code highly inefficient.

The main improvement would be on some of the code organization, including putting some functionality into functions so the code is clearer and more organized.
Some parts of the code were repeated, for example getting the user input/option, so that would be a great improvement point!
Adding new options to the current code could become confusing, as the error handling would have to be changed/done manually, and could affect the cycle of other parts of the code.

## Testing

I could not find any big flaws for the application testing. It tested a lot of options that could be used. The tests covered a lot of the options. They were also organized into different tests and sections, which is great.
Nothing to add to this part!

Thisisnotgoingtowork:
#### Nice work on the app! 
I tried it and run it and it worked pretty well!

## Code Design
Overall, there aren't any huge issue, the code is pretty straightforward and understandable, there are just some minor suggestions:
1. In the FinalApplication.cpp, the Menu logic is using nested while and switch logic more than thousands of lines, this may lead to hurting the readability and testability of the app. In this case, maybe you could create small command function per operation and dispatch through a table of lambdas? In that way I think the app is more stable.
2. Again, in your FinalApplication.cpp, there are something like #include "ExpressionParser.cpp" which is risking ODR violation, you may want to change it to header files. 
3. GetColumnIndex, GetDataValue declare default stream parameters in both header and source. I think you do want to check that as well. 
 

## Use of Modern C++
1. In the MainMenu function, using something like (filename.ends_with(".csv")) as raw string is not really the best way, you could add some paths check at this moment like using the std::filesystem::path and path.extension()==".csv" for example
2. There are some redundant of the FinalApplication object in the main.cpp, Either keep MainMenu static and call it directly or make methods non-static and use the instance consistently.
3. Function ReferenceVector keeps raw pointers, but menu code dereferences without null check, which might lead to breaking down the functions. You might want to check them again?


After all, I tried the app, it is pretty interesting to use, and the menus are very detailed. They could help a lot.

Group-05/FinalApplication.cpp
owenhaiar: ^ Seconded, see https://www.cse.msu.edu/~cse450/Emplex.html

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1526 @@
...
+      int index =
+          GetColumnIndex(static_cast<int>(std::get<1>(grid.Shape())), os, is);
+
+      os << "\nBelow is the grid: \n";
+      grid.Print(os);
+
+      os << "\nMean at column " << index << ": " << grid.ColumnMean(index)
+         << std::endl;
+      return;
+    } else if (option == "cmed") {
```
owenhaiar, line 457: Slightly out of scope for my review topic, but you have a lot of redundant else keywords in this file. If you return in the if block above, these else-ifs can just be ifs which makes it clearer that you're returning

Group-05/FinalApplication.h
```diff
@@ -0,0 +1,69 @@
...
+ * columns or the entire grid
+ * - Use comparison operations (less than, less than or equal to, greater than,
+ * greater than or equal to, equal to, and not equal to) to compare a given
+ * value against values in a column
+ *
+ * When done, users can export the grid into a .csv file.
+ */
+class FinalApplication {
+ private:
+  [[nodiscard]] static std::optional<double> IsValidDouble(
```
owenhaiar, line 33: Good use of static helpers

Group-05/src/ExpressionParser.h
```diff
@@ -3,103 +3,106 @@
...
-#include <functional>
 
-namespace cse {
+#include "Datum.h"
+
+using cse::Datum;
 
+namespace cse {
+template <typename SymbolTableType>
+concept HasIndexing = requires(SymbolTableType t, size_t i) {
```
owenhaiar, line 28: Good use of concepts

Group-05/src/ExpressionParser.h
```diff
@@ -3,103 +3,106 @@
...
   /**
    * @brief set char to subtraction operator
-   * 
+   *
    * @param sub_char comparison char
    */
-  constexpr void subtraction(char& sub_char){
-    sub_char='-';
-  }
+  constexpr void subtraction(char &sub_char) { sub_char = '-'; }
```
owenhaiar, line 48: All the math ops can be made static

Group-05/FinalApplication.h
```diff
@@ -0,0 +1,69 @@
...
+ * - Print the grid
+ * - Calculate the mean, median, standard deviation, mode, min, and max of
+ * columns or the entire grid
+ * - Use comparison operations (less than, less than or equal to, greater than,
+ * greater than or equal to, equal to, and not equal to) to compare a given
+ * value against values in a column
+ *
+ * When done, users can export the grid into a .csv file.
+ */
+class FinalApplication {
```
owenhaiar, line 31: Most of these functions can be marked noexcept

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1526 @@
...
+#include "src/ExpressionParser.cpp"
+#include "src/ExpressionParser.h"
+
+/**
+ * Checks if a string can be converted into a double
+ * @param test_string The string to check
+ * @return an optional double if the string can be converted, empty optional if
+ * not
+ */
+std::optional<double> FinalApplication::IsValidDouble(
```
owenhaiar, line 54: The parameter names here and in a few other functions dont match the names in the declaration.

owenhaiar:
Overall, the app is straightforward and simple to use. I thought the procedural style with lots of static functions made sense for the project specs.

# Implementation Quality
The code compiles without warnings. However, the compilation command you gave does not work on my machine due to the ReferenceVector header file being included in the command. The code compiles fine when you remove this file. It would be helpful to create a makefile for the project.

The biggest issue in the implementation was the large functions that handle menu output, input checking, then perform logic based on input. The menu outputs could be made const strings, outputted by a function. The input checking could also be function that returns an enum for the option type, rather than a string in some places and an int in others. Additionally, some menu options are single characters, with others are ints, while others are strings, which makes the code more confusing to read and the app slightly more confusing to use. Enums for options would also make it easier to add more options in the future.

The trim lambda is duplicated twice in GridMenu and CreateGridMenu, and there is also a trim whitespace function in CSVfile.

Overall the efficiency was good, as nothing too computationally complex is being done. A lexer could be used for parsing the user inputs, rather than regex searches. This could also make your user input more intuitive.

# Use of Modern C++
Good use of std::optional for input checking.

consexpr, static, and [[nodiscard]] were used where appropriate. Many of the functions could be marked noexcept.

Very minor issues: The auto keyword could be used more often, and in some places you manually specified template arguments when the compiler could do it automatically. You are also missing const from some local variables. 


I think the project is well done, and the functions are designed and organized well at a high level, but the code has a lot of unnecessarily complex control flow and code that needs to be moved to helper functions.

chrispaul-lab:
### Testing & Error Checking

_General Review_ 
Overall the test cases and error checking was quite strong. It was robust and especially covered a lot of edge case testing. ExpressionParserTests could benefit from adding a few more tests regarding parsing errors and some edge cases. Some of the tests benefit from a few tests regrading performance under stress. Also, there are some tests which verify that exceptions are thrown with `CHECK_THROWS`, but don't verify the specific exception message or type in all cases.

I have a few comments for each test file (no large fixes, small details for improvement) that I have added in-line. 

All in all, good work!


### User Experience
The UX is well built and intuitive, but sometimes there are few improvements I can suggest, which are mainly focused on some navigational friction. Besides the points below, you all built a great app! Very powerful features!

**Key Strengths:** 
- Great use of error handling (such as the out-of-bounds column indices being caught)
- Immediate responsiveness from CLI 
- Flexible grid creation & import/export are very handy features
- Input validation is a great 

**Potential Improvements**
- The menu navigation can sometimes be a little bit tedious. For example, actions that should be simple (i.e editing a cell) requires 3 -4 levels of menu selection. 
- After completing some operations (especially math functions), I got kicked back to the main menu. It might be a little easier to kick back into a contextually based menu. 
- This is a smaller note, but it might be a little easier for users if there is a consistent input type. Currently, there are letters for the main menu, numbers for manipulation menu, and text commands for math menu. 
- Adding some headers and maybe borders might be useful for larger grids
- String and numeric values are also not visually aligned at all times, which can make the table messy at certain times
- It would be very nice to have a little more guidance in Equation Builder. 
- Some of the most powerful features (such as the Equation Builder) are buried under layers of menu's to reach.
- Extending off the last suggestion, overall it might be beneficial to allow for direct command input (i,e `edit 1,2 45`)
- Undo/redo functionality might be very useful
- Batch editing could also be a very useful feature
- A status bar which displays the current grid dimensions and filename at all times could be super helpful while going through the UX.

Group-05/FinalApplication.cpp
MaxKrawec: We actually removed this lambda. It wasn't needed.

Group-05/main.cpp
MaxKrawec: Thanks

Group-05/FinalApplication.cpp
MaxKrawec: Interesting. The regex satisfies what we need for this check, but thanks for pointing this out! I'll consider it for future implementations.

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1526 @@
...
+ * @param max_index The maximum index for the grid
+ * @param os ostream used for output and testing
+ * @param is istream used for input and testing
+ * @param label optional label for row or column, defaults to column
+ * @return A valid column (or row) index for the grid
+ */
+int FinalApplication::GetColumnIndex(int max_index,
+                                     std::ostream &os = std::cout,
+                                     std::istream &is = std::cin,
+                                     const std::string &label) {
```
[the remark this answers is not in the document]
MaxKrawec: I greatly refactored almost every function, enhancing simplicity and readability. The code is in a much better state now. Thanks!

Group-05/FinalApplication.cpp
MaxKrawec: I greatly refactored almost every function, enhancing simplicity and readability. The code is in a much better state now. Thanks!

tests/Group-05/FinalApplicationTests.cpp
ShahaabAli: Thank you!

Group-05/FinalApplication.h
MaxKrawec: Thanks

Group-05/FinalApplication.cpp
ShahaabAli: Great suggestion! We just implemented it.

Group-05/FinalApplication.cpp
ShahaabAli: Thank you for the suggestion. We now have a function that validates the input, and split this function into a few sub-functions to make the code more readable.

Group-05/FinalApplication.cpp
ShahaabAli: This will definetly improve code readability. We have now implemented this.

Group-05/FinalApplication.cpp
ShahaabAli: We tried to implement this, but after doing so, it made the code harder to read. It is a great thought though. Thank you!

Group-05/FinalApplication.cpp
ShahaabAli: We tried to implement this, but after doing so, it made the code harder to read. It is a great thought though. Thank you!

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1526 @@
...
+    os << "Enter your choice: ";
+    std::string input;
+    std::getline(is, input);
+    std::optional<int> input_int = IsValidInt(input);
+
+    if (input_int.has_value()) {
+      choice = input_int.value();
+      switch (choice) {
+        case 1:
+          PrintSubmenu(grid, os, is);
```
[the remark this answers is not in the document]
ShahaabAli: We tried to implement this, but after doing so, it made the code harder to read. It is a great thought though. Thank you!

Group-05/FinalApplication.cpp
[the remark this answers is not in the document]
ShahaabAli: All of these parameters now have default values, and are listed in the .h file. Thank you!

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1526 @@
...
+  }
+  return std::nullopt;
+}
+
+/**
+ * Checks if a string can be converted into an int
+ * @param test_string The string to check
+ * @return an optional int if the string can be converted, empty optional if not
+ */
+std::optional<int> FinalApplication::IsValidInt(
```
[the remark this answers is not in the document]
ShahaabAli: Good suggestion, thank you!

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1526 @@
...
+      for (const double &mode_val : grid.ColumnMode(index)) {
+        os << mode_val << " ";
+      }
+      os << std::endl;
+      return;
+    } else if (option == "sum") {
+      os << "\nBelow is the grid: \n";
+      grid.Print(os);
+
+      auto summary = grid.CalculateDataGridMathSummary();
```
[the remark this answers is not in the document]
ShahaabAli: This will improve readability!!! Thank you!

Group-05/FinalApplication.cpp
```diff
@@ -0,0 +1,1526 @@
...
+  }
+}
+
+/**
+ * Prompts the user for a Datum value
+ * @param os ostream used for output and testing
+ * @param is istream used for input and testing
+ * @return A datum value
+ */
+cse::Datum FinalApplication::GetDataValue(std::ostream &os = std::cout,
```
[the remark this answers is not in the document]
ShahaabAli: Thank you

Group-05/FinalApplication.cpp
ShahaabAli: Updated!

Group-05/FinalApplication.cpp
MaxKrawec: When we refactored the code, we removed those 10,000 values, so I don't believe this is necessary any more. Thank you for the suggestion tho!

Group-05/FinalApplication.cpp
MaxKrawec: Added. Thanks

MaxKrawec:
We created a new PR from our main application branch --> main. Please review and use that one. However, this PR contains the code reviews.

The new PR: https://github.com/CSE498/CSE498-Spring2025/pull/256

MaxKrawec:
We created a new PR from our main application branch --> main. Please review and use that one. However, this PR contains the code reviews.

The new PR: https://github.com/CSE498/CSE498-Spring2025/pull/256

PR #241, opened 2025-04-22
Bellman Final Application

Devon-FoxElster:
Final Application: 
(Command line operated application to interface with BruteForceOptimizer in order to solve Knapsack Problems)

How to Run: Navigate to Group-10/Application and run make. Prompts and help messages will lead from there, but reach out with any questions

Test files have been provided to run through the solver.

Advanced Class commits are included as they were needed for the application


EvilJuiceBox (commented):
I see you have a number of .txt files that are launcher files (?) for some predefined inputs of your program to demonstrate the classes.

I would suggest instead either wrapping them into a readme.md file that outlines some example use case with inputs for someone who wants to test the classes, or build a general purpose launcher that allows you to input some predefined scripts. You can avoid a little bit of clutter in the directory.

tests/Group-10/TestComboManager.cpp
EvilJuiceBox, line 1: The test cases for this class are much better. Still could test a number of different inputs, but generally OK.

tests/Group-10/TestComboManager.cpp
```diff
@@ -1,293 +1,500 @@
...
+  do {
+    std::vector<double> combo = cm.GetCurrentCombo();
+    INFO("Forward Combo " << count << ": " << (combo.empty() ? "(empty)" : ""));
+    count++;
+  } while (cm.NextCombo());
+  REQUIRE(count == expectedTotal);
 }
 
 // Test 4: Combination size 0 (should yield one empty combination).
 TEST_CASE("ComboManager: Combination size 0", "[ComboManager]") {
```
EvilJuiceBox, line 94: Good job on this edge case.

Should also test negative items here for completeness.

tests/Group-10/TestComboManager.cpp
```diff
@@ -1,293 +1,500 @@
...
-    
-    std::vector<char> combo = cm.GetCurrentCombo();
-    INFO("Combo contents: ");
-    REQUIRE(combo.size() == letters.size());
-    for (size_t i = 0; i < letters.size(); ++i) {
-        REQUIRE(combo[i] == letters[i]);
-    }
-    // No next combination should exist.
-    REQUIRE_FALSE(cm.NextCombo());
+TEST_CASE("ComboManager: Combination size equal to container size",
```
EvilJuiceBox, line 109: Good test, try above allowed size as well

tests/Group-10/TestComboManager.cpp
```diff
@@ -1,293 +1,500 @@
...
+  REQUIRE(first != second);
+
+  // Move back to the first combination.
+  moved = cm.PrevCombo();
+  REQUIRE(moved);
+  std::vector<int> firstAgain = cm.GetCurrentCombo();
+  REQUIRE(first == firstAgain);
+}
+
+// Test 13: Lexicographical order of generated combinations
```
EvilJuiceBox, line 254: Nice test

ShahaabAli (commented):
User Experience Feedback:

When the application is first run, an explanation of what the Knapsack problem is, and how your application solves it, would make the purpose of your application more clear. This could also be implemented as a standalone menu option. Personally, I think it would be best to output this information when the application is first run.

A more detailed explanation or example of how the CSV file should be formatted would be very helpful if the user wanted to use custom CSV data as an input. This could also be something that is outputted when the application is first run, or another menu option to explain csv format. Maybe a '-h' command could be utilized to get this information.

'set-capacity' has a '-h' option to show flags, but there are no flag options aside from '-h'

When running the brute-force command with the -c flag, the following message, Total time to compute: 0 milliseconds, can be made more clear by mentioning which method is used to get that time (optimized or brute-force). Also, the the time should be adjusted based on significant figures.

I think the commands can better be explained, so they are easier to understand, especially for a user who is not very familiar with the knapsack problem. For example, instead of, show-capacity - Shows the default capacity (if one is set), say, show-capacity - Shows the default Knapsack capacity (if one is set). This is just my personal opinion, up to you if you think this makes it more clear.

Outputting the items when using brute-force as comma + space separated, instead of new line separated may make it easier to read, especially if there are many items.

I think the brute-force command could be better named to clarify its purpose, and a flag can be chosen to choose how the problem is solved, with the default being brute-force. For example, it could be named 'solve'.

Overall, the user interface is well made. After implementing the above changes, I think it will be even more clear to the end user to better interact with your application. Great Job!

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,387 @@
...
+  std::string token;
+
+  while (std::getline(ss, token, delimiter)) {
+    sections.push_back(token);
+  }
+
+  return sections;
+}
+
+// Utility function to measure execution time
```
ShahaabAli, line 70: Use doxygen comments for style consistency

ShahaabAli (commented):
Code Design Feedback:

Overall, you did a great job. I love how you modularized a lot of the code, and abstracted the implementation for tokenizing, error handling, etc. Here is my feedback for things that can be improved.

Some functions are missing doxygen style comments, this would be great to improve code readability. Additionally, having more comments in function bodies would also help clarify how the code works.

Consider renaming some functions to be a little more clear. Some functions are named generically, and are unclear as to what their purpose is. For example, split. I would name it tokenize or splitstring. This will also help avoid namespace collisions because you will be avoiding common function names.

Instead of using global variables to store application settings, consider creating a class that stores the settings. An object of this class can be initialized at the beginning of the program and passed as needed. The reason I say this, is because this can be a cause of side effects and potential bugs.

Everything else looks great to me. Good Job Again!

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,387 @@
...
+      auto solutionPair = optimizer.FindOptimalSolution();
+      PrintOptimizerResults(solutionPair);
+    });
+    if (settings.timeSearch || settings.compare) PrintTiming(optimizedTime);
+  }
+
+  if (settings.compare) {
+    // Print comparison results
+    std::cout << "Speedup: "
+              << ((unoptimizedTime / optimizedTime) *
```
IvanBega, line 204: You initialize optimizedTime with 0.0 and then divide by this value without checking if it is not 0. Is it possible for optimizeTime to remain 0.0 until this line of code?

Group-10/Application/ApplicationTools.hpp
```diff
@@ -0,0 +1,72 @@
...
+struct OptimizerSettings {
+  std::string filename = "";
+  std::vector<cse::Item> itemList = {};
+  double capacity = 0.0;
+  bool multipleRepeats = false;
+  bool weightless = false;
+  bool compare = false;
+  bool optimized = false;
+  bool timeSearch = false;
+  std::optional<double> defaultCapacity = std::nullopt;
```
IvanBega, line 70: Good example of safe code

tests/Group-10/TestFinalApp.cpp
```diff
@@ -0,0 +1,259 @@
+#include <fstream>
```
IvanBega, line 1: A lot of work had been done to test the final app. Multiple use-cases were considered. Maybe some additional test cases can be added for files that exist but have incorrect structure

tests/Group-10/TestFinalApp.cpp
```diff
@@ -0,0 +1,259 @@
...
+TEST_CASE("Repeats mode computes correct total value from script3",
+          "[FinalApp][Repeats]") {
+  std::ifstream inputFile("script3.txt");
+  std::ostringstream capturedOutput;
+
+  // Redirect std::cout to capture output
+  auto* originalBuf = std::cout.rdbuf();
+  std::cout.rdbuf(capturedOutput.rdbuf());
+
+  if (inputFile.is_open()) {
```
IvanBega, line 189: Consider replacing this structure with a standard catch2 tool like REQUIRE(inputFile.is_open()) for clarity and readability

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,387 @@
...
+#include "../Classes/ComboManager.hpp"
+#include "../Classes/StaticVector.hpp"
+#include "ApplicationTools.hpp"
+
+static cse::OptimizerSettings settings;
+static const cse::PresetMessages messages;
+static constexpr std::size_t CAPACITY_ARGLENGTH = 10;
+static constexpr double THRESHOLD_TO_SECONDS = 500;
+static constexpr double MILLISECONDS_IN_SECONDS = 1000;
+static constexpr double SPEEDUP_ADJUSTMENT_CONSTANT = 100;
```
IvanBega, line 24: Global variables can be wrapped in a namespace to avoid name collisions in the future. This will also improve code safety by avoiding bugs caused by variable overshadowing

IvanBega (commented):
** Final Application Review **

Testing and Error Checking

I started this part by running the app and trying different kinds of input to crash the program. I wasn't able to do that. However, once I slightly modified an example3.txt file by removing one value from a single line and ran the program again (brute-force example3.txt), I immediately got a segmentation fault. So adding more checks, in addition with negative test cases can be added. Apart from that, all other incorrect inputs were caught and handled appropriately

Implementation Quality

In general, code appears to be written safe, efficient, and flexible. I would say that new features can be easily added on top of existing ones without major refactoring. It is understandable that some shortcuts had to be taken due to lack of time, like usage of global variables etc. However, I still think that application.hpp can be refactored just a little bit to make the code safer

Lilryesx:
Code Design
Overall, group 10’s code design is clean, modular, and very logical across major components like BruteForceOptimizer, ComboManager, MemoFunction, MemoryFactory, PermutationManager, and StaticVector.  I approached my review by simply reviewing each file and testing myself to understand what was happening without reading the comments. For the most part, I was able to do so successfully. Overall, class responsibilities are clearly separated, and file organization makes it easy to follow how different parts fit together. The use of clear public APIs with minimal clutter in headers is well done, and dependencies between files are kept lightweight, which helps with understanding and maintainability.

One improvement would be adding more detailed documentation comments at the top of each class and above important methods, especially to explain assumptions (like whether MemoFunction is thread-safe, whether StaticVector checks for overflow, or whether ComboManager allows duplicate combinations - examples from ChatGPT because I couldn't think of any lol). Small naming tweaks could help too, such as making methods like next() more specific (nextCombo() or nextPermutation()) to avoid confusion if a project grows larger. Edge cases like overflow handling or thread-safety could also be clarified either in comments or by adding simple runtime checks.

Overall, the design is very strong — logical, easy to navigate, and well-structured — and would be even more polished and beginner-friendly with just a few added comments and tiny naming improvements.

Y'all did amazing!!!!!!

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+
+void PrintTerminal() { std::cout << "\033[32m\nBellman-Application> \033[0m"; }
+
+std::string RedError(std::string &&message) {
+  return "\033[31m" + message + "\033[0m\n";
+}
+
+[[noreturn]] void EndProgram() { std::exit(0); }
+
+void PrintOptimizerResults(std::pair<double, std::vector<cse::Item>> solution) {
```
132ikl, line 96: I understand using a `std::pair` for your solution as the return value of `FindOptimalSolution` instead of making a whole new struct, but it's a bit clearer here if you separate out the solution value and items and separate parameters to this function. I would recommend changing the signature to something like `void PrintOptimizerResults(double optimalValue, std::vector<cse::Item>> const &solutionItems)`, and then using destructuring when you call `FindOptimalSolution`, so you can do

```cpp
auto [optimalValue, solutionItems] = optimizer.FindOptimalSolution();
PrintOptimizerResults(double optimalValue, solutionItems);
```

Group-10/Application/application.cpp
132ikl: An assert would probably be good here

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+    ResetGlobalVariables();
+    // split input
+    //  command -> text file seperated by a space
+    auto arguments = split(input, ' ');
+    auto argMgr = CreateArgManager(arguments);
+    if (!arguments.size()) {
+      PrintTerminal();
+      continue;
+    } else if ((arguments.at(0) == "q" || arguments.at(0) == "Q" ||
+                arguments.at(0) == "quit" || arguments.at(0) == "exit")) {
```
132ikl, line 259: You check `arguments.at(0)` a bunch of times here and below, you could pull this out into a variable above and then check against that instead

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+            std::cout << messages.capacityHelp;
+            PrintTerminal();
+            continue;
+          }
+          double givenCapacity = std::stod(arguments.at(1));
+          settings.defaultCapacity = givenCapacity;
+          settings.capacity = givenCapacity;
+          std::cout << settings.defaultCapacity.value()
+                    << " set as default capacity.\n";
+        } catch (const std::exception &e) {
```
132ikl, line 297: You can specify this as `std::invalid_argument` to only catch the `std::stod` exception, and not other unrelated exceptions

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+          std::cout << RedError("**Invalid value given for the capacity");
+        }
+      }
+    }
+
+    else if (arguments[0] == "show-capacity") {
+      (settings.defaultCapacity.has_value())
+          ? std::cout << "Default value set to: "
+                      << settings.defaultCapacity.value()
+          : std::cout << "No default value set.";
```
132ikl, line 307: Why use a ternary here instead of a normal `if`? Ternaries are normally useful when you want to use the result in an expression, but in this case you're just looking for the side-effect of calling `std::cout`

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+        } catch (const std::exception &e) {
+          std::cout << RedError("**Invalid value given for the capacity");
+        }
+      }
+    }
+
+    else if (arguments[0] == "show-capacity") {
+      (settings.defaultCapacity.has_value())
+          ? std::cout << "Default value set to: "
+                      << settings.defaultCapacity.value()
```
132ikl, line 306: Nitpick, but you should use the `*` operator on an optional (`*settings.defaultCapacity`) after you've already checked if it has a value using `has_value`. The `value` method has a built-in check, similar to `std::vector`'s `at`, which is redundant if you've already checked it. This is similar to doing something like this:

```cpp
std::vector<int> myVector{1, 3, ..., 4};
if (myVector.size() > 5)
  return myVector.at(5); // the `.at` here is redundant, since we've already checked that the size is greater than 5
```

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+      if (arguments.size() < 2) {
+        std::cout << RedError(
+            "**Please specify a filename as the second argument.");
+        PrintTerminal();
+        continue;
+      }
+      settings.filename = arguments[1];
+
+      if (!settings.filename.contains(".txt") &&
+          !settings.filename.contains(".csv")) {
```
132ikl, line 328: You should use `ends_with` here. I could have `my_data.csv.gz`!

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+      } else {
+        std::cout << RedError("** \""+settings.filename+"\" is not a valid file path");
+        PrintTerminal();
+        continue;
+      }
+      std::string valString;
+      std::string toFind = "-capacity=";
+      auto capacityArg = std::find_if(
+          arguments.begin(), arguments.end(),
+          [=](auto str) { return str.find(toFind) != std::string::npos; });
```
132ikl, line 357: You can break this with
```
brute-force test-capacity=.csv
```

You should use `str.starts_with` here instead. Also, I don't think your lambda (or the parameter) need to be pass-by-value?

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+  settings.itemList = {};
+  settings.capacity = (settings.defaultCapacity.has_value())
+                          ? settings.defaultCapacity.value()
+                          : 0.0;
+  settings.multipleRepeats = false;
+  settings.weightless = false;
+  settings.compare = false;
+  settings.optimized = false;
+  settings.timeSearch = false;
+}
```
132ikl, line 50: Why not remove this function and default (re-)initialize settings?

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+#include <vector>
+
+#include "../../Group-06/ArgManager/ArgManager.cpp"
+#include "../../Group-06/CommandLine/CommandLine.cpp"
+#include "../Classes/BruteForceOptimizer.hpp"
+#include "../Classes/ComboManager.hpp"
+#include "../Classes/StaticVector.hpp"
+#include "ApplicationTools.hpp"
+
+static cse::OptimizerSettings settings;
```
132ikl, line 19: Your application logic should _probably_ be in a class and put this inside of there instead of using a global variable, but I won't tell anyone if you keep it like this 🤫

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "../../Group-06/ArgManager/ArgManager.cpp"
+#include "../../Group-06/CommandLine/CommandLine.cpp"
+#include "../Classes/BruteForceOptimizer.hpp"
+#include "../Classes/ComboManager.hpp"
+#include "../Classes/StaticVector.hpp"
```
132ikl, line 16: Consider adding these directories to your include directories in your Makefile, so you can just do `#include "BruteForceOptimizer.hpp"` instead of referring to where those files exist on the filesystem.

Group-10/Classes/BruteForceOptimizer.hpp
```diff
@@ -0,0 +1,222 @@
...
+  Item() = default;
+  Item(std::string newName, double newWeight, double newValue)
+      : value(newValue), weight(newWeight), name(newName) {};
+  ~Item() = default;
+
+  /**
+   * @brief comparison operator for two Items
+   * @return indication of the comparison between two Items
+   */
+  auto operator<=>(const Item&) const = default;
```
132ikl, line 40: Love the default spaceship operator usage

Group-10/Classes/BruteForceOptimizer.hpp
```diff
@@ -0,0 +1,222 @@
...
+ */
+struct Item {
+  double value;
+  double weight;
+  std::string name;
+
+  Item() = default;
+  Item(std::string newName, double newWeight, double newValue)
+      : value(newValue), weight(newWeight), name(newName) {};
+  ~Item() = default;
```
132ikl, line 34: These shouldn't be necessary, you're not really doing anything fancy and the aggregate initializer should already exist for this struct. It's a matter of taste, but if you're going to use a `struct` rather than a `class` it's nice to keep it as slim as possible (but honestly it's more vibes than anything else)

Group-10/Classes/BruteForceOptimizer.hpp
```diff
@@ -0,0 +1,222 @@
...
+                           double currentValue) {
+    assert(currentWeight >= 0 && currentValue >= 0);
+
+    auto updateScore = [this](double currentValue) {
+      bestScore_ = currentValue;
+      bestSelection_ = currentSelection_;
+    };
+
+    auto scoreCheck = [this, currentWeight, currentValue,
+                       updateScore]() mutable {
```
132ikl, line 129: Does this need to be mutable? It doesn't look like you're setting anything inside

Group-10/Classes/BruteForceOptimizer.hpp
```diff
@@ -0,0 +1,222 @@
...
+    sortedItems_ = items_;
+    std::stable_sort(sortedItems_.begin(), sortedItems_.end(),
+                     std::greater<Item>());
+    // Create value array based on the remaining overall value at any index
+    double totalValue = 0.0;
+    for (auto iter = sortedItems_.rbegin(); iter != sortedItems_.rend();
+         ++iter) {
+      if (repeating_) {
+        int maxItemAmount = capacity_ / (*iter).weight;
+        totalValue += (maxItemAmount * (*iter).value);
```
132ikl, line 211: I would recommend using the `->` operator here instead of dereferencing followed by a member access

Group-10/Classes/BruteForceOptimizer.hpp
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * @brief Stores the list of items in the optimizer.
+   *
+   * @param newItems The vector of items to manage.
+   */
+  void SetItems(const std::vector<Item>& newItems) {
+    items_ = newItems;
+    if (optimizeEnabled_) OptimizationPreprocessing_();
+  }
```
132ikl, line 83: Shouldn't you also clear `bestSelection_`, `currentSelection_`, `scoreTracker_`, and `sortedItems_` here?

Group-10/Classes/BruteForceOptimizer.hpp
```diff
@@ -0,0 +1,222 @@
...
+    for (auto iter = sortedItems_.rbegin(); iter != sortedItems_.rend();
+         ++iter) {
+      if (repeating_) {
+        int maxItemAmount = capacity_ / (*iter).weight;
+        totalValue += (maxItemAmount * (*iter).value);
+      } else {
+        totalValue += (*iter).value;
+      }
+      scoreTracker_.push_back(totalValue);
+    }
```
132ikl, line 216: You could use a range here if you wanted to be fancy. Some combination of `views::reverse`, `views::transform`, and `views::to` should do the trick.

Group-10/Classes/MemoryFactory.hpp
```diff
@@ -25,6 +24,12 @@ namespace cse {
 template <typename Object>
 class MemoryFactory {
  private:
+  /// @brief Size used when none is provided by the user
+  static constexpr int DEFAULT_SIZE = 10;
+
+  /// @brief Multiplier used when the MemoryFactory is allocating new space
+  static constexpr double EXPANSION_SIZE = 1.618;
```
132ikl, line 31: A classic choice! If you are interested in this, I highly recommend this video that goes way more into depth about this than I ever thought I would be interested in: https://youtu.be/GZPqDvG615k

132ikl (commented):
Very nice application! I love your command-line interface. I also really like how you can compare the unoptimized search and the optimized search. I may have gotten a bit carried away with my comments, but I hope you find them helpful :smile:

Group-10/Classes/BruteForceOptimizer.hpp
Devon-FoxElster: Most of these are covered in other areas, but I can see why it would be cleaner to include them here as well

If it is set to be optimized it will automatically reset scoreTracker_ and sortedItems_ in the OptimizationPreprocessing_ functions

currentSelection_ gets cleared every time we run the FindOptimalSolution function so it is already being reset any time we want to get an answer from the class

bestSelection wasn't being cleared (woops) but with bestScore being reset in the FindOptimalSolution function it was automatically wiped once the first score was checked. I'm adding a clear for bestSelection just to be thorough though.

Group-10/Classes/BruteForceOptimizer.hpp
Devon-FoxElster: Thanks for the catch! I won't lie I forgot about that operator when I wrote it lol

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+  std::cout << std::endl;
+  for (const auto &item : vector) {
+    std::cout << item << "," << std::endl;
+  }
+  std::cout << std::endl;
+}
+
+void PrintTerminal() { std::cout << "\033[32m\nBellman-Application> \033[0m"; }
+
+std::string RedError(std::string &&message) {
```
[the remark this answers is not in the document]
Devon-FoxElster: Thanks!

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+cse::ArgManager CreateArgManager(std::vector<std::string> &args) {
+  std::vector<char *> argV(args.size());
+  for (std::size_t i = 0; i < args.size(); ++i) {
+    // argV[i] = strdup(args[i].c_str()); // duplicates a new string. Must use
+    // because c_str returns a const string
+    argV[i] = args[i].data();
+  }
+  cse::ArgManager mgr(args.size(), argV.data());
+  return mgr;
+}
```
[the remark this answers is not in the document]
Devon-FoxElster: This was actually a topic of debate when we were working with the class, but ultimately we decided it would be simpler in this situation to make an interface function ourselves. Otherwise we would need to coordinate with their group to get a changed pushed to our branch while they were trying to work on their own application. A similar issue now, where class is wrapping up in less than a week.

In a long-term view I agree that it would be easier to have that built into the class itself, and I appreciate the feedback!

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+ * vector must outlive the argmanager, which should be fine for our code.abort
+ * @param args the vector of strings
+ * @return an argManager object
+ */
+cse::ArgManager CreateArgManager(std::vector<std::string> &args) {
+  std::vector<char *> argV(args.size());
+  for (std::size_t i = 0; i < args.size(); ++i) {
+    // argV[i] = strdup(args[i].c_str()); // duplicates a new string. Must use
+    // because c_str returns a const string
+    argV[i] = args[i].data();
```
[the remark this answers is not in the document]
Devon-FoxElster: I'm not entirely certain on `c_str` and `data` doing the same thing here, as I just switched them out as a test and the comment seems to be accurate? The function definition reads `c_str`'s return as constexpr const char* while `data` is just constexpr char *.

Group-10/Application/application.cpp
Devon-FoxElster: Implemented

Group-10/Application/application.cpp
Devon-FoxElster: Implemented

Group-10/Application/application.cpp
Devon-FoxElster: It felt more readable, not much else to it. Either implementation would work. I'll switch it over to an if block

tests/Group-10/TestBruteForceOptimizer.cpp
```diff
@@ -1,88 +1,245 @@
...
+      measureTime([&]() { optimizer.FindOptimalSolution(); });
+
+  auto optimizedScore = optimizer.FindOptimalSolution().first;
+  std::cout << "\n13 Item Test:\n";
+  std::cout << "Unoptimized Time: " << unoptimizedTime << std::endl;
+  std::cout << "Optimized Time: " << optimizedTime << std::endl;
+  std::cout << "Speedup: "
+            << ((unoptimizedTime - optimizedTime) / unoptimizedTime) * 100
+            << "%\n";
+  REQUIRE(unoptimizedTime > optimizedTime);
```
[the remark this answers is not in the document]
Muelleralex1021: Removed Testing for Speed-up times

Group-10/Application/application.cpp
Devon-FoxElster: For the majority of them you're 100% correct, it was mostly the need to keep default capacity and capacity constant throughout usage of the program that led to the partial reset. I ended up implementing a mix of your suggestion and the current code where the settings object gets reset, and we just confirm that capacity carries over as it is supposed to.

Group-10/Application/application.cpp
Bcrims: Fixed, good find!

Group-10/Application/ApplicationTools.hpp
```diff
@@ -0,0 +1,72 @@
...
+      "considerations. The algorithm will only "
+      "consider the number of items in respect to the capacity\n"
+      "\033[32m-repeats\033[0m (\033[32m-r\033[0m) = Allows for items to be "
+      "used "
+      "multiple times in the "
+      "solution. (WARNING: May drastically increase solving time)\n";
+};
+
+struct OptimizerSettings {
+  std::string filename = "";
```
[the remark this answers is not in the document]
Devon-FoxElster: Filename can't be empty in any area where filename is accessed, if that makes sense. One of our checks forces a filename if the user is trying to use the brute-force command. The empty string initialization was just intended to make sure it starts out empty.

Group-10/Application/application.cpp
Devon-FoxElster: Partially implemented, with the timing structure we have set up we aren't able to deconstruct inline. I did change the PrintOptimizerResults signature though

Group-10/Application/ApplicationTools.hpp
```diff
@@ -0,0 +1,72 @@
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "../Classes/BruteForceOptimizer.hpp"
+
+namespace cse {
+struct PresetMessages {
```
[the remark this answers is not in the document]
Devon-FoxElster: Implemented, thanks for the tip!

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+    std::cout << item << "," << std::endl;
+  }
+  std::cout << std::endl;
+}
+
+void PrintTerminal() { std::cout << "\033[32m\nBellman-Application> \033[0m"; }
+
+std::string RedError(std::string &&message) {
+  return "\033[31m" + message + "\033[0m\n";
+}
```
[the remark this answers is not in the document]
Devon-FoxElster: Implemented

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+          settings.itemList = ConstructItems(textFile);
+        } catch (std::length_error e) {
+          std::cout << RedError(e.what());
+          PrintTerminal();
+          continue;
+        } catch (std::invalid_argument e) {
+          std::cout << RedError(e.what());
+          PrintTerminal();
+          continue;
+        }
```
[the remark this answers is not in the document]
Devon-FoxElster: Resolved, made them references rather than values

tests/Group-10/TestFinalApp.cpp
Devon-FoxElster: Checking if the file was open was more for extra error-checking, since it wouldn't actually test anything in our application we are hesitant to add it to a REQUIRE() block. We'll keep it as is, but seeing as we made the input files ourselves this won't trip regardless.

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+  while (itemIter != chosenItems.end()) {
+    auto item = (*itemIter).first;
+    auto itemCount = (*itemIter).second;
+    std::cout << item;
+    if (itemCount > 1) {
+      std::cout << '(' << itemCount << "x)";
+    }
+    std::cout << std::endl;
+    ++itemIter;
+  }
```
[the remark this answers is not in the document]
Devon-FoxElster: Implemented after a fashion, the original idea was to prevent that tailing comma but it got tossed out in favor of having items on different lines. I combined this section with the old PrintVector function to handle outputting the solution vector.

Group-10/Application/application.cpp
```diff
@@ -0,0 +1,417 @@
...
+        std::to_string(itemData.size()));
+    }
+    if (!isDigit(itemData[1])) {
+      throw std::invalid_argument("CSV parse error on line " + std::to_string(lineNumber) + " invalid argument.");
+    }
+    if (!isDigit(itemData[2])) {
+      throw std::invalid_argument("CSV parse error on line " + std::to_string(lineNumber) + " invalid argument.");
+    }
+    
+    cse::Item item(itemData[0], std::stod(itemData[1]), std::stod(itemData[2]));
```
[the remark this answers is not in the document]
Devon-FoxElster: Implemented

Group-10/Application/application.cpp
Devon-FoxElster: Implemented

Group-10/Application/application.cpp
Devon-FoxElster: Implemented

tests/Group-10/TestBruteForceOptimizer.cpp
```diff
@@ -1,88 +1,245 @@
...
-        REQUIRE(result.first == 6.0);
-        double sumWeight = 0.0;
-        for (auto const& item : result.second)
-            sumWeight += item.weight;
-        REQUIRE(sumWeight == 6.0);
-    }
+  SECTION("Mixed example, capacity=6") {
+    cse::BruteForceOptimizer optimizer;
+    optimizer.SetItems(
+        {{"A", 2.0, 2.0}, {"B", 2.0, 2.0}, {"C", 6.0, 6.0}, {"D", 3.0, 3.0}});
```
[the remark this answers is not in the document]
Bcrims: Test cases have been added for both scenarios regarding weight exceeding value

tests/Group-10/TestBruteForceOptimizer.cpp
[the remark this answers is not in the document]
Bcrims: Added test cases with much more variation and random assignment

Group-10/Application/application.cpp
Devon-FoxElster: 🤫

Group-10/Application/application.cpp
Devon-FoxElster: Implemented!

Group-10/Classes/BruteForceOptimizer.hpp
Devon-FoxElster: The reasoning behind this was that it could give us a little more flexibility in how we ordered the actual variables vs. how we constructed them, as std::greater goes in order from top to bottom but we preferred having name first in the actual constructor for ease of reading

PR #253, opened 2025-04-29
Liskov Small Bug Fixes

merryymary:
- Fixed WebLayout Testing (no index file existed to inject into)
- Added tests for additional functions in weblayout
- Fixed image size not saving correctly to json after being resized
- Added more doc comments


owenhaiar (approved):
Big fan

graceKC04 (approved):

PR #257, opened 2025-04-30
Updated testcases based on Event id now being string. 

graceKC04:
Updated application classes based on feedback.


merryymary (approved):
10/10

